PR #4, opened 2020-10-16
Fix typo Workflow -> Activity

robholland:


antstorm (approved):
Nice catch, thank you!

PR #5, opened 2020-10-17
Add a rake task to generate GRPC files.

robholland:


Rakefile
```diff
@@ -0,0 +1,6 @@
+desc 'Generate ruby files from protocol buffers'
+task :gen do
```
antstorm, line 2: nit: `gen:proto` ? Also consider using `Makefile` instead in case we need to expose `Rakefile` to the users of the gem

Rakefile
```diff
@@ -0,0 +1,6 @@
...
+  Dir['proto/temporal/**/*.proto'].each do |file|
```
antstorm, line 3: gRPC tpols accepts wildcard:

```
bundle exec grpc_tools_ruby_protoc -I proto --ruby_out=lib/gen --grpc_out=lib/gen proto/temporal/api/**/*.proto
```

Rakefile
robholland: `Makefile` sounds good, will fix.

Rakefile
robholland: I think we'd need to quote the string with the `**` to avoid the shell trying to parse it but I'll give it a go.

antstorm: This is the command I used to generate the current protos, so it does work

robholland:
Switched to a Makefile using the same structure as `proto/Makefile`.

antstorm (approved):

PR #7, opened 2020-10-17
Rename domain -> namespace to match temporal.

robholland:


antstorm:
@robholland this PR makes sense, but I have some concerns. Mainly this will make cadence <> temporal gems non-syncable, which will make developing both a bit of a nightmare. Can we keep modifications to the client only and use aliases (supporting both `domain` and `namespace`) in all public APIs?

antstorm (approved):
Also looking good and needs a rebase

antstorm:
@robholland I think this one's ready to go after a rebase

robholland:
Rebased.

antstorm (approved):

PR #13, opened 2020-10-20
Time related fixes.

robholland:
Use Google::Protobuf::Timestamp methods for time conversion.

Remove _seconds suffixes from some attributes.


antstorm (commented):
Looks good, needs rebase

robholland:
Updated.

lib/temporal/activity/task_processor.rb
```diff
@@ -48,7 +48,9 @@ def process
       attr_reader :task, :domain, :task_token, :activity_name, :activity_class, :client, :middleware_chain
 
       def queue_time_ms
-        ((task.started_timestamp - task.scheduled_timestamp_of_this_attempt) / 1_000_000).round
+        scheduled = task.current_attempt_scheduled_time.to_f
+        started = task.started_timestamp.to_f
+        ((started - scheduled) * 1_000).round
```
antstorm, line 53: There should be exactly the same logic in `lib/temporal/workflow/decision_task_processor.rb`, can you please apply these changes to it as well?

lib/temporal/workflow/execution_info.rb
```diff
@@ -26,8 +26,8 @@ def self.generate_from(response)
           workflow: response.type.name,
           workflow_id: response.execution.workflowId,
           run_id: response.execution.run_id,
-          start_time: Utils.time_from_nanos(response.startTime),
-          close_time: Utils.time_from_nanos(response.closeTime),
+          start_time: response.start_time.to_time,
```
antstorm, line 29: As far as I understand `response.start_time` is an instance of `Google::Protobuf::Timestamp`, which does not implement `#to_time`

antstorm (changes requested):

lib/temporal/workflow/execution_info.rb
robholland: It's defined here I believe: https://github.com/protocolbuffers/protobuf/blob/master/ruby/lib/google/protobuf/well_known_types.rb#L74

antstorm:
```ruby
>> timestamp = Google::Protobuf::Timestamp.new
>> timestamp.to_time

NoMethodError (undefined method `to_time' for <Google::Protobuf::Timestamp: seconds: 0, nanos: 0>:Google::Protobuf::Timestamp)
```

robholland: Hmm. Will dig into that to understand why that's not present.

antstorm: Had to explicitly `require 'google/protobuf/well_known_types'` to make it work I don't think it is loaded by default

robholland: Yep, found an issue discussing that. Updated my PR to require it.

lib/temporal/activity/task_processor.rb
robholland: Done.

antstorm (approved):
👍 

Let's also start adding up the specs where these are missing. Will save us from these small mistakes moving forward

lib/temporal/workflow/serializer/schedule_activity.rb
```diff
@@ -43,7 +43,6 @@ def serialize_retry_policy(retry_policy)
             maximum_interval: retry_policy.max_interval,
             maximum_attempts: retry_policy.max_attempts,
             non_retriable_error_reasons: non_retriable_errors,
-            expiration_interval: retry_policy.expiration_interval
```
antstorm, line 46: Hmm, this change is totally unrelated and questionable (https://temporalio.slack.com/archives/CTRCR8RBP/p1603714682311600), can we leave it out of this PR?

robholland: The field no longer exists, I can split it into another PR if you'd prefer but it needs to be removed.

robholland: Removed from this PR.

antstorm (approved):

PR #14, opened 2020-10-20
Remove rbi file which isn't required.

robholland:


antstorm:
@robholland this is required for those using Sorbet. Any good reason for removing it?

robholland:
Ah, my apologies I misremembered what Sorbet was and thought it was a tool related to producing bindings for Thrift or similar. The file needs updating if it's staying as it refers to constants no longer present.

PR #16, opened 2020-10-27
Spec fixes.

robholland:
Most specs now passing: 205 examples, 12 failures


lib/temporal/workflow/execution_info.rb
```diff
@@ -22,12 +22,12 @@ class ExecutionInfo < Struct.new(:workflow, :workflow_id, :run_id, :start_time,
...
         new(
           workflow: response.type.name,
-          workflow_id: response.execution.workflowId,
+          workflow_id: response.execution.workflow_id,
           run_id: response.execution.run_id,
           start_time: response.start_time.to_time,
           close_time: response.close_time.to_time,
-          status: response.closeStatus || RUNNING_STATUS,
-          history_length: response.historyLength,
+          status: response.status.to_s.delete_prefix('WORKFLOW_EXECUTION_STATUS_').to_sym,
```
robholland, line 29: Not sure about this @antstorm ?

spec/fabricators/grpc/activity_task_fabricator.rb
```diff
@@ -0,0 +1,15 @@
+require 'securerandom'
+
+Fabricator(:api_activity_task, from: Temporal::Api::WorkflowService::V1::PollActivityTaskQueueResponse) do
+  transient :task_token, :activity_name, :headers
+
+  activity_id { SecureRandom.uuid }
+  task_token { |attrs| attrs[:task_token] || SecureRandom.uuid }
+  activity_type { Fabricate(:api_activity_type) }
+  input Temporal::Workflow::Serializer::Payload.new('').to_proto
```
robholland, line 9: Not sure about using the serializer in a test, maybe it's fine. Might want to use the verbose Api structure or a test specific helper though?

robholland:
Wish it showed diffs for the moved files rather than a delete/add, makes it harder to review.

lib/temporal/testing/temporal_override.rb
```diff
@@ -64,7 +64,7 @@ def start_locally(workflow, *input, **args)
         run_id = SecureRandom.uuid
 
         if !allowed?(workflow_id, reuse_policy)
-          raise TemporalThrift::WorkflowExecutionAlreadyStartedError,
+          raise Temporal::Api::ErrorDetails::V1::WorkflowExecutionAlreadyStartedFailure,
```
robholland, line 67: This is no longer an exception class so not sure what you'd like to raise instead here.

robholland:
`205 examples, 8 failures`

lib/temporal/testing/temporal_override.rb
antstorm: We'll have to define all these errors in the clients and then adopt Cadence to raise the same ones. This way we'll have a strong separation between the API and the core logic

lib/temporal/workflow/execution_info.rb
antstorm: Probably worth adding a mapping between API statuses and our own ones. At least that way you can be sure about the statuses returned to developers

spec/fabricators/grpc/activity_task_fabricator.rb
antstorm: I think nested Fabricator might be a better options here similar to `activity_type`

spec/spec_helper.rb
```diff
@@ -15,6 +15,8 @@
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 
 require 'temporal'
+require 'google/protobuf/well_known_types'
```
antstorm, line 18: Are we using it outside of the grpc_client specs (where it is already included)?

spec/unit/lib/temporal_spec.rb
```diff
@@ -1,16 +1,17 @@
 require 'temporal'
 require 'temporal/workflow'
+require 'pry'
```
antstorm, line 3: Happy to include it as a spec dependency, but not in a particular spec file

spec/unit/lib/temporal_spec.rb
```diff
@@ -1,16 +1,17 @@
...
 
 describe Temporal do
   describe 'client operations' do
     let(:client) { instance_double(Temporal::Client::GRPCClient) }
 
     before { allow(Temporal::Client).to receive(:generate).and_return(client) }
-    after { described_class.remove_instance_variable(:@client) }
+    after { described_class.remove_instance_variable(:@client) rescue NameError }
```
antstorm, line 10: Hmm… why `rescue NameError`?

antstorm (changes requested):
Few comments, but mostly looks good

spec/spec_helper.rb
robholland: We're using it to generate Google::Protobuf::Timestamps in the fabricators.

spec/unit/lib/temporal_spec.rb
robholland: Oops, meant to remove this. I may stick it spec_helper though.

spec/unit/lib/temporal_spec.rb
robholland: This is because some specs don't actually generate a client because they stub a response which prevents `generate` ever being called. In these cases removing the instance variable throws a NameError because it doesn't exist. I can either rescue NameError or I can check if the ivar is present before trying to remove, no real preference. I did this because it's shorter.

spec/unit/lib/temporal_spec.rb
robholland: Done.

robholland:
All specs passing.

spec/unit/lib/temporal_spec.rb
antstorm: Makes sense 👍

antstorm (approved):
Outstanding job, thank you! There are few fixes that I'll address directly on `master`, let's merge this one before it gets too big

PR #19, opened 2020-10-31
Rename Decision to WorkflowTask.

robholland:


lib/temporal.rb
```diff
@@ -50,16 +50,16 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
       )
     end
 
-    def reset_workflow(namespace, workflow_id, run_id, decision_task_id: nil, reason: 'manual reset')
-      decision_task_id ||= get_last_completed_decision_task(namespace, workflow_id, run_id)
-      raise Error, 'Could not find a completed decision task event' unless decision_task_id
+    def reset_workflow(namespace, workflow_id, run_id, workflow_task_id: nil, reason: 'manual reset')
+      workflow_task_id ||= get_last_completed_workflow_task(namespace, workflow_id, run_id)
+      raise Error, 'Could not find a completed workflow_task task event' unless workflow_task_id
```
antstorm, line 55: nit: `workflow_task task`

robholland: Fixed.

lib/temporal/workflow/context.rb
```diff
@@ -41,7 +41,7 @@ def execute_activity(activity_class, *input, **args)
 
         execution_options = ExecutionOptions.new(activity_class, options)
 
-        decision = Decision::ScheduleActivity.new(
+        workflow_task = WorkflowTask::ScheduleActivity.new(
```
antstorm, line 44: I think this is a wrong assumption — top level `Decision` is not the same as `DecisionTask`, so therefore it does not translate to `WorkflowTask`. This should be a `Command`, which is what a "decision" was used as a synonym of

lib/temporal/workflow/history/event.rb
```diff
@@ -42,8 +42,8 @@ def initialize(raw_event)
         end
 
         # Returns the ID of the first event associated with the current event,
-        # referred to as a "decision" event. Not related to DecisionTask.
-        def decision_id
+        # referred to as a "workflow_task" event. Not related to WorkflowTaskTask.
```
antstorm, line 45: ^ This comment explain exactly what is wrong with the rename :)

`decision_id` should be `command_id`

antstorm (changes requested):

robholland:
Lets just scrap this, I'll file a fresh PR :)

PR #21, opened 2020-11-04
Wrap some GRPC errors.

robholland:
Don't leak GRPC errors outside the client.

More work to do here, this is just low hanging fruit.

### DIFF
--- a/lib/temporal/client/grpc_client.rb
+++ b/lib/temporal/client/grpc_client.rb
@@ -31,6 +31,8 @@ module Temporal
           )
         )
         client.register_namespace(request)
+      rescue GRPC::AlreadyExists => e
+        raise Temporal::NamespaceAlreadyExistsFailure, e.details
       end
 
       def describe_namespace(name:)
@@ -97,6 +99,10 @@ module Temporal
         end
 
         client.start_workflow_execution(request)
+      rescue GRPC::AlreadyExists => e
+        # Feel like there should be cleaner way to do this...
+        run_id = e.details[/RunId: (.*)\.$/, 1]
+        raise Temporal::WorkflowExecutionAlreadyStartedFailure.new(e.details, run_id)
       end
 
       def get_workflow_execution_history(namespace:, workflow_id:, run_id:)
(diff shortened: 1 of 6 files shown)


lib/temporal/client/grpc_client.rb
```diff
@@ -97,6 +99,10 @@ def start_workflow_execution(
         end
 
         client.start_workflow_execution(request)
+      rescue GRPC::AlreadyExists => e
+        # Feel like there should be cleaner way to do this...
+        run_id = e.details[/RunId: (.*)\.$/, 1]
+        raise Temporal::WorkflowExecutionAlreadyStartedFailure.new(e.details, run_id)
```
antstorm, line 105: I actually wonder if it's helpful to have `run_id` provided here

spec/unit/lib/temporal/testing_spec.rb
```diff
@@ -84,15 +84,15 @@ def execute; end
             let(:status) { Temporal::Workflow::ExecutionInfo::RUNNING_STATUS }
 
             it 'raises error' do
-              expect { subject }.to raise_error(error_class)
+              expect { subject }.to raise_error(error_class) { |e| expect(e.run_id).to eql run_id }
```
antstorm, line 87: nit: would suggest using parenthesis around `run_id`

lib/temporal/client/grpc_client.rb
robholland: It's useful for singleton workflows where you wish to ensure the workflow is started but it may already be running. Without having `run_id` returned as part of the exception you would have to then send another request to find the `run_id` if you wish to send a signal or so forth.

```
def self.start
  Temporal.start_workflow(
    ...
    options: { workflow_id: 'my-workflow' }
   )
rescue Temporal::WorkflowExecutionAlreadyStartedFailure => e
  e.run_id
end
```

antstorm (approved):
Looks good 👍

lib/temporal/client/grpc_client.rb
antstorm: Ah, good point 👍

PR #23, opened 2020-11-04
Silence the logger during tests.

robholland:


antstorm (changes requested):
Can you please move the changes to a file under `spec/config/`? These are all loaded and is a better way of configuring spec behaviour rather than putting everything in the spec_helper

robholland:
Done.

antstorm (approved):
👍

PR #24, opened 2020-11-04
Bring retries inline with temporal docs.

robholland:


antstorm (commented):
I have mixed feelings about this change. In itself it bears little to no value since there isn't a requirement to match dev-facing names with Temporal's API

This wasn't identical to Cadence API in the first place, so I wonder if there's any reason for going this way

robholland:
It makes it easier to look up what the values actually do in the temporal docs. I don't feel strongly about it though.

antstorm:
@robholland let's postpone this one then. Will revisit a bit later

PR #25, opened 2020-11-04
Allow use of dry-types 1.4.0

drewhoskins-stripe:
### Summary
It was convenient for us to use `dry-types` v1.4.0 .
It seems to have caused us to start enforcing stricter typing between symbols and strings, at least for hash keys.
```
E, [2020-11-04T11:10:50.686328 #69212] ERROR -- temporal_client: Workflow task for LocalHelloWorldWorkflow failed with: #<TypeError: wrong argument type Symbol (expected String)>
D, [2020-11-04T11:10:50.686395 #69212] DEBUG -- temporal_client: /Users/drewhoskins/stripe/temporal-ruby/lib/temporal/workflow/serializer/record_marker.rb:12:in `initialize'
```

### Test Plan
Set up a Docker with temporal_server on my localhost (see instructions https://docs.temporal.io/docs/install-temporal-server/)
```
cd examples
bundle install
./bin/worker
./bin/trigger LocalHelloWorldWorkflow
./bin/trigger ParentWorkflow
./bin/trigger TripBookingWorkflow 7
```
This last one looks like an error is expected:
```
Saga execution aborted: #<Trip::MakePaymentActivity::InsufficientFunds: Unable to charge 183.2>
```


drewhoskins-stripe:
(will resolve conflicts)

antstorm (approved):
The change looks good, thank you Drew 👍 

One note: the dry-* dependencies were removed from `cadence-ruby` after forking this gem out — https://github.com/coinbase/cadence-ruby/pull/12/files. I think it makes sense to do this here as well, since it's not a hard requirement but is only needed if you want to type your activity/workflow inputs

drewhoskins-stripe:
Nice, that'd be a welcome change from our perspective as well.  I'll see if I have time to tackle that.

antstorm:
@drewhoskins-stripe I've got some time, so will merge this and remove the dependency. Thanks a lot for contributing 👍

PR #28, opened 2020-11-05
Handle nil payloads.

robholland:
This happens if the Payloads object is missing, rather than present with
an empty payloads array.

### DIFF
--- a/lib/temporal/workflow/state_manager.rb
+++ b/lib/temporal/workflow/state_manager.rb
@@ -305,7 +305,7 @@ module Temporal
       end
 
       def parse_payload(payload)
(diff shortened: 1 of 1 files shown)


antstorm:
What is the case for missing the `Payloads` object? It should be serialized even with actual input being nil — `lib/temporal/client/serializer/payload.rb:11`

robholland:
We receive protobufs with `nil` for payloads from the Temporal server. Triggered easily with `tcl workflow start ....`

antstorm:
@robholland got it, looks like Temporal is serializing `nil`s slightly differently

antstorm (approved):
👍

PR #31, opened 2020-12-17
Make middleware logging clearer.

immacodethat-rob:
Previously it was a bit ambigious whether the workflow as a whole was finished versus a single task.


antstorm (approved):
Looks great, thank you for clarifying it

PR #33, opened 2021-01-05
Make execute_in_context clean up after itself

drewhoskins-stripe:
Minor change to improve hygiene.
Previously, the change to the thread-local context persisted on the main thread in testing mode, making it a bit harder to isolate tests that used `Workflow.execute_locally` from one another.

Test plan:
`rspec ./spec/unit/lib/temporal/testing_spec.rb`
Also ran end-to-end tests in non-testing mode in our environment.

### DIFF
--- a/spec/unit/lib/temporal/testing_spec.rb
+++ b/spec/unit/lib/temporal/testing_spec.rb
@@ -35,6 +35,21 @@ describe Temporal::Testing::TemporalOverride do
       Temporal::Testing.local! { example.run }
     end
 
+    describe 'Workflow.execute_locally' do
+      let(:workflow) { TestTemporalOverrideWorkflow.new(nil) }
+
+      before { allow(TestTemporalOverrideWorkflow).to receive(:new).and_return(workflow) }
+
+      it 'restores original context after finishing' do
+        allow(workflow).to receive(:execute)
+
+        TestTemporalOverrideWorkflow.execute_locally
+
+        expect(workflow).to have_received(:execute)
+        expect(Temporal::ThreadLocalContext.get).to equal(nil)
+      end
+    end
+
     describe 'Temporal.start_workflow' do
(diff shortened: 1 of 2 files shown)


spec/unit/lib/temporal/testing_spec.rb
```diff
@@ -35,6 +35,21 @@ def execute; end
       Temporal::Testing.local! { example.run }
     end
 
+    describe 'Workflow.execute_locally' do
+      let(:workflow) { TestTemporalOverrideWorkflow.new(nil) }
+
+      before { allow(TestTemporalOverrideWorkflow).to receive(:new).and_return(workflow) }
+
+      it 'restores original context after finishing' do
```
antstorm, line 43: Can you please add a failure case example where an error is raised?

spec/unit/lib/temporal/testing_spec.rb
```diff
@@ -35,6 +35,21 @@ def execute; end
       Temporal::Testing.local! { example.run }
     end
 
+    describe 'Workflow.execute_locally' do
```
antstorm, line 38: I think this test belongs in `spec/unit/lib/temporal/workflow_spec.rb`

antstorm (changes requested):
@drewhoskins-stripe thanks for submitting this. Few small comments, but looks good overall 👍

spec/unit/lib/temporal/testing_spec.rb
drewhoskins-stripe: While the fix is in code that's also used in production, the method I'm calling is in `lib/temporal/testing`.  LMK if you still want me to move it.
Also LMK if you think I should test a non-testing path.  I didn't bother because it doesn't feel like this really comes up in production since each workflow looked to have its own thread so there wouldn't be context corruption.

antstorm (approved):
Looks good 👍

PR #35, opened 2021-01-06
Allow local testing of Temporal.schedule_workflow

drewhoskins-stripe:
# Summary

Today if you call `Temporal.schedule_workflow` in local testing mode, you'll get an RPC error.
Scenario: someone is exercising some code that schedules a workflow in an integration test. They should be able to check that the workflow was scheduled with the cron parameters they expect, then run it to simulate deferred execution.

Seems harder to test/simulate the actual cron schedule, so  I didn't do that.

# Test plan
`rspec ./spec/unit/lib/temporal/testing_spec.rb`


antstorm:
This is a very good addition, thank you. One thing that I'm not so sure about is allowing the execution of the scheduled workflows in tests. I just can't think of a scenario when this would be needed.

Normally when you are scheduling a workflow you never expect it to be run immediately (min resolution of a cron schedule is a minute). So when you are testing the `schedule_workflow` you only really care that it got scheduled (which you got covered via `scheduled` hash). Then when you need to test the workflow's code itself — you can do this as you would normally using the `start_workflow` call.

Please let me know if I'm missing something here

lib/temporal/testing.rb
```diff
@@ -23,6 +25,31 @@ def local?
...
+        scheduled_executions[workflow_id].call
+      end
+
+      def execute_all_scheduled_workflows
+        scheduled_executions.transform_values(&:call)
+      end
+
+      # Populated by Temporal.schedule_workflow
+      # format: { <workflow_id>: <cron schedule string>, ... }
+      def schedules
```
antstorm, line 44: I might be good to have schedules live close to `executions` in the override module. One particular option I'm considering is being able to check if a workflow has been scheduled using Temporal's public API. I wonder how does that actually work today — will Temporal allow you to query a workflow that was scheduled, but not yet executed…

drewhoskins-stripe:
@antstorm good question.  It's not useful in unit tests, but it comes up for us in integration tests.  Suppose an API call schedules a workflow (among other things).  You'd like to make sure the API did the right thing (i.e. scheduled it properly, called it with the right parameters, etc).  So you just execute exactly what it enqueued, and that's a better test than separately re-creating what it did.

lib/temporal/testing.rb
drewhoskins-stripe: Interesting.  It seems weird to put schedules and scheduled_executions  in different modules, but I could move them both back to the overrides.    (See my most recent stacked commit)
The reason I moved was for sorbet; I needed to add types to our `.rbi` files and I didn't want to add these types to the main `Temporal` namespace even though they aren't currently available in production.
Maybe we could move them over if/when the production version becomes available?  WDYT?

anthonydmitriyev:
@drewhoskins-stripe but in that case you are kinda trying to just execute the workflow at the time of schedule, there's no real benefit in waiting to execute it later. I would also expect that inline workflow execution can be quite expensive, so kinda creates an anti-pattern if you just need to check that something got scheduled with the right arguments.

I wonder if there's a better way to check argument compatibility

lib/temporal/testing.rb
anthonydmitriyev: Right, sorry, my bad. Definitely don't want to extend `Temporal`'s interface with testing methods, should be identical to the real production one. Makes sense to use `Temporal::Testing`. However it feels like the `Temporal::Testing` should call into some common private class to get the scheduled workflows and executions.

One of the smells here is essentially having publicly writable `Temporal::Testing.schedules` which can be updated from the test code directly.

Basically we need something to track schedules/executions that's neither `Temporal` nor `Temporal::Testing`, but used by both

drewhoskins-stripe:
@anthonydmitriyev Today with our legacy system, our test wrappers around API calls execute the immediately scheduled async events and don't execute the deferred ones.  This works reasonably well in practice, allowing the developer to choose whether to execute is a flexible option, and we haven't found that giving developers the option of executing the async events has led to anti-patterns.  (More often, they're surprised when stuff doesn't get run that they wanted to be run.)
Auto-executing everything inline isn't a great option for speed as you point out.  Synchronous execution also reorders the typical execution flow, making the test less valid.
As a typical example, you could imagine say 30 tests testing an API endpoint, and only 2 will care about the workflow's result enough to execute it and pay the cost in CI.

That said, I like the idea of having an opt-in inline execution mode for scheduled workflows.  I may not have time to build that in this PR, but it could be easily added!

lib/temporal/testing.rb
drewhoskins-stripe: Agreed, will move into a private class.

lib/temporal/testing.rb
```diff
@@ -43,6 +44,30 @@ def with_mode(new_mode, &block)
         @mode = previous_mode
       end
     end
+
+    # When Temporal.schedule_workflow is called in a test in local mode, we defer the execution and do
+    # not do it automatically.
+    # You can execute them or inspect their cron schedules using this module.
+    module ScheduledWorkflows
```
anthonydmitriyev, line 51: this needs to be in its own file

lib/temporal/testing/scheduled_workflows.rb
```diff
@@ -0,0 +1,55 @@
+module Temporal
+  module Testing
+    class WorkflowIDNotScheduled < ClientError; end
+
+    # Implementation for Temporal::Testing::ScheduledWorkflwows
```
anthonydmitriyev, line 5: This is not really providing implementation for `ScheduledWorkflow` does it? As it there's nothing connecting the two (even the interface is different), so maybe call it `ScheduledWorkflowStore` or something?

lib/temporal/testing/scheduled_workflows.rb
```diff
@@ -0,0 +1,55 @@
...
+    module ScheduledWorkflowsImpl
+      class << self
+
+        # For someone who wants to assert that the schedule is what they expect.
+        # Populated by Temporal.schedule_workflow
+        # format: { <workflow_id>: <cron schedule string>, ... }
+        def schedules
+          _schedules.dup.freeze
```
anthonydmitriyev, line 13: This will create a new object every time it is accessed, which seems wasteful. Why not do this in a functional way instead — create a new frozen hash every time you add a value?

lib/temporal/testing/scheduled_workflows.rb
```diff
@@ -0,0 +1,55 @@
...
+        end
+
+        def add(workflow_id:, cron_schedule:, executor_lambda:)
+          _schedules[workflow_id] = cron_schedule
+          _scheduled_executions[workflow_id] = executor_lambda
+        end
+
+        private def _schedules
```
anthonydmitriyev, line 21: what does "_" signify?

lib/temporal/testing/scheduled_workflows.rb
```diff
@@ -0,0 +1,55 @@
...
+          @schedules ||= {}
+        end
+
+        private def _scheduled_executions
+          @scheduled_executions ||= {}
+        end
+
+        # Populated by Temporal.schedule_workflow
+        def scheduled_executions
```
anthonydmitriyev, line 30: this is not being used, is it?

spec/unit/lib/temporal/testing_spec.rb
```diff
@@ -30,6 +37,65 @@ def execute; end
...
+  describe 'In local testing mode, Temporal.schedule_workflow' do
+    around do |example|
+      Temporal::Testing.local! { example.run }
+    end
+
+    it 'allows the test to simulate deferred executions' do
+      workflow = TestTemporalOverrideWorkflow.new(nil)
+      allow(TestTemporalOverrideWorkflow).to receive(:new).and_return(workflow)
+      workflow2 = TestTemporalOverride2Workflow.new(nil)
+      allow(TestTemporalOverride2Workflow).to receive(:new).and_return(workflow2)
```
anthonydmitriyev, line 49: You can simply:

```ruby
workflow = TestTemporalOverrideWorkflow.new(nil)
workflow2 = TestTemporalOverrideWorkflow.new(nil)

allow(TestTemporalOverrideWorkflow).to receive(:new).and_return(workflow, workflow2)
```

avoiding the need for a second blank workflow, which doesn't really do anything

spec/unit/lib/temporal/testing_spec.rb
```diff
@@ -30,6 +37,65 @@ def execute; end
     end
   end
 
+  describe 'In local testing mode, Temporal.schedule_workflow' do
```
anthonydmitriyev, line 40: nit: to fit with the other specs —

```ruby
describe 'Temporal.scheduled_workflow' do
  context 'in local testing mode' do
    ...
  end
end
```

lib/temporal/testing/scheduled_workflows.rb
```diff
@@ -0,0 +1,55 @@
...
+            raise Temporal::Testing::WorkflowIDNotScheduled,
+              "There is no workflow with id #{workflow_id} that was scheduled with Temporal.schedule_workflow.\n"\
+              "Options: #{_scheduled_executions.keys}"
+          end
+
+          _scheduled_executions[workflow_id].call
+        end
+
+        def execute_all
+          _scheduled_executions.transform_values(&:call)
```
anthonydmitriyev, line 50: nit: is this intentionally transforming the values? Or did you mean something like `_scheduled_executions.values.each(&:call)`?

lib/temporal/testing/scheduled_workflows.rb
```diff
@@ -0,0 +1,75 @@
...
+      # format: { <workflow_id>: <cron schedule string>, ... }
+      def self.cron_schedules
+        Private::Store.schedules
+      end
+
+      def self.clear_all
+        Private::Store.clear_all
+      end
+
+      module Private
```
drewhoskins-stripe, line 28: You can make a class/module private with `private_constant`, but that doesn't work using it from two modules.  This is a convention we use in our codebase (though we have tooling to enforce it.)
Open to better options.

lib/temporal/testing/scheduled_workflows.rb
drewhoskins-stripe: (removed it)

antstorm (approved):
Looks good, thank you for addressing my comments 👍

PR #36, opened 2021-01-14
Make execution timeout default to infinite

drewhoskins-stripe:
## Description
See the comment for justification.

## Test Plan
Enqueued a workflow with no override, went to temporal web UI:
Screen Shot 2021-01-13 at 10 38 14 AM

### DIFF
--- a/lib/temporal/configuration.rb
+++ b/lib/temporal/configuration.rb
@@ -6,8 +6,11 @@ module Temporal
     attr_reader :timeouts
     attr_accessor :client_type, :host, :port, :logger, :metrics_adapter, :namespace, :task_queue, :headers
 
+    # We want an infinite execution timeout for cron schedules and other perpetual workflows.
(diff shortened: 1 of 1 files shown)


antstorm (approved):
Looks good 👍

lib/temporal/configuration.rb
```diff
@@ -6,8 +6,11 @@ class Configuration
     attr_reader :timeouts
     attr_accessor :client_type, :host, :port, :logger, :metrics_adapter, :namespace, :task_queue, :headers
 
+    # We want an infinite execution timeout for cron schedules and other perpetual workflows.
+    # We choose an 10-year execution timeout because that's the maximum the cassandra DB supports,
+    # matching the go SDK, see https://github.com/temporalio/sdk-go/blob/d96130dad3d2bc189bc7626543bd5911cc07ff6d/internal/internal_workflow_testsuite.go#L68
```
antstorm, line 11: nit: Probably not the best reference since it's coming from a test suite and not production code

PR #37, opened 2021-01-26
Ensure we use a clean RunId.

robholland:
Previously we would accidentally catch the text following the run id.

Reported by: @drewhoskins-stripe


anthonydmitriyev (changes requested):
Looks good, but let's add a spec to make sure we don't regress later if someone decides to change that RegExp further

drewhoskins-stripe:
Thanks!

antstorm (approved):
Brilliant, thank you!

PR #39, opened 2021-02-04
Add an error callback.

robholland:
This is useful to report exceptions to an exception monitoring service.

### DIFF
--- a/lib/temporal/activity/poller.rb
+++ b/lib/temporal/activity/poller.rb
@@ -69,6 +69,11 @@ module Temporal
         client.poll_activity_task_queue(namespace: namespace, task_queue: task_queue)
       rescue StandardError => error
         Temporal.logger.error("Unable to poll activity task queue: #{error.inspect}")
+
+        Temporal.configuration.error_handlers.each do |handler|
+          handler.call(error, nil)
+        end
+
         nil
       end
 
--- a/lib/temporal/activity/poller.rb
+++ b/lib/temporal/activity/poller.rb
@@ -2,6 +2,7 @@ require 'temporal/client'
 require 'temporal/thread_pool'
 require 'temporal/middleware/chain'
 require 'temporal/activity/task_processor'
+require 'temporal/error_handler'
 
 module Temporal
   class Activity
--- a/lib/temporal/activity/poller.rb
+++ b/lib/temporal/activity/poller.rb
@@ -70,7 +70,7 @@ module Temporal
       rescue StandardError => error
         Temporal.logger.error("Unable to poll activity task queue: #{error.inspect}")
 
-        Temporal::ErrorHandler.handle(error, nil)
+        Temporal::ErrorHandler.handle(error)
 
         nil
       end
--- a/lib/temporal/activity/poller.rb
+++ b/lib/temporal/activity/poller.rb
@@ -70,9 +70,7 @@ module Temporal
       rescue StandardError => error
         Temporal.logger.error("Unable to poll activity task queue: #{error.inspect}")
 
-        Temporal.configuration.error_handlers.each do |handler|
-          handler.call(error, nil)
-        end
+        Temporal::ErrorHandler.handle(error, nil)
 
         nil
       end
--- a/lib/temporal/activity/task_processor.rb
+++ b/lib/temporal/activity/task_processor.rb
@@ -7,8 +7,6 @@ require 'temporal/json'
 module Temporal
   class Activity
     class TaskProcessor
-      attr_accessor :metadata
-
       def initialize(task, namespace, activity_lookup, client, middleware_chain)
         @task = task
         @namespace = namespace
@@ -50,7 +48,7 @@ module Temporal
 
       private
 
-      attr_reader :task, :namespace, :task_token, :activity_name, :activity_class, :client, :middleware_chain
+      attr_reader :task, :namespace, :task_token, :activity_name, :activity_class, :client, :middleware_chain, :metadata
 
       def queue_time_ms
         scheduled = task.current_attempt_scheduled_time.to_f
--- a/lib/temporal/activity/task_processor.rb
+++ b/lib/temporal/activity/task_processor.rb
@@ -1,4 +1,5 @@
 require 'temporal/metadata'
+require 'temporal/error_handler'
 require 'temporal/errors'
 require 'temporal/activity/context'
 require 'temporal/json'
--- a/lib/temporal/activity/task_processor.rb
+++ b/lib/temporal/activity/task_processor.rb
@@ -30,7 +30,7 @@ module Temporal
 
         context = Activity::Context.new(client, metadata)
 
-        result = middleware_chain.invoke(metadata, task) do
+        result = middleware_chain.invoke(metadata) do
           activity_class.execute_in_context(context, parse_payload(task.input))
         end
 
--- a/lib/temporal/activity/task_processor.rb
+++ b/lib/temporal/activity/task_processor.rb
@@ -38,9 +38,7 @@ module Temporal
       rescue StandardError, ScriptError => error
         respond_failed(error)
 
-        Temporal.configuration.error_handlers.each do |handler|
-          handler.call(error, metadata)
-        end
+        Temporal::ErrorHandler.handle(error, metadata)
       ensure
         time_diff_ms = ((Time.now - start_time) * 1000).round
         Temporal.metrics.timing('activity_task.latency', time_diff_ms, activity: activity_name)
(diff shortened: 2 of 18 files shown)


robholland:
I stubbed out all of the history/state machine innards as setting up fabricators fully enough for all that code to function was complicated and didn't seem to be worth it for the contexts I wanted to test.

lib/temporal/activity/poller.rb
```diff
@@ -66,6 +66,11 @@ def poll_for_task
         client.poll_activity_task_queue(namespace: namespace, task_queue: task_queue)
       rescue StandardError => error
         Temporal.logger.error("Unable to poll activity task queue: #{error.inspect}")
+
+        Temporal.configuration.error_handlers.each do |handler|
```
antstorm, line 70: Let's get this under a more concise interface, something like `Temporal::ErrorHandler.handle(error, nil)`

lib/temporal/activity/poller.rb
```diff
@@ -66,6 +66,11 @@ def poll_for_task
...
+          handler.call(error, nil)
```
antstorm, line 71: We'll also need to put guards around call to the provided lambdas, we can't afford crashing on buggy error handling

lib/temporal/configuration.rb
```diff
@@ -3,7 +3,7 @@
 
 module Temporal
   class Configuration
-    attr_reader :timeouts
+    attr_reader :timeouts, :error_handlers
```
antstorm, line 6: Let's add a method for adding these handlers rather than expose the whole array. Something like:

```ruby
def on_error(&block)
  @error_handlers << block
end
```

lib/temporal/workflow/poller.rb
```diff
@@ -60,6 +60,11 @@ def poll_for_task
         client.poll_workflow_task_queue(namespace: namespace, task_queue: task_queue)
       rescue StandardError => error
         Temporal.logger.error("Unable to poll workflow task queue: #{error.inspect}")
+
+        Temporal.configuration.error_handlers.each do |handler|
```
antstorm, line 64: Not sure about this one… or maybe we need to have multiple handler types

antstorm (changes requested):

lib/temporal/workflow/poller.rb
robholland: We need to be able to report WorkflowNotRegistered.

lib/temporal/workflow/poller.rb
```diff
@@ -60,6 +60,9 @@ def poll_for_task
         client.poll_workflow_task_queue(namespace: namespace, task_queue: task_queue)
       rescue StandardError => error
         Temporal.logger.error("Unable to poll workflow task queue: #{error.inspect}")
+
+        Temporal::ErrorHandler.handle(error, nil)
```
antstorm, line 64: I think this is expected to fail on a polling timeout, so I don't think it's super helpful to report. Maybe we need to rescue different errors here to distinguish expected failure

lib/temporal/workflow.rb
```diff
@@ -19,6 +19,8 @@ def self.execute_in_context(context, input)
       Temporal.logger.error("Workflow execution failed with: #{error.inspect}")
       Temporal.logger.debug(error.backtrace.join("\n"))
 
+      Temporal::ErrorHandler.handle(error, context.send(:metadata))
```
antstorm, line 22: We should probably make `metadata` public, `.send` suggests that we're doing something wrong here

antstorm (commented):
Any ideas why the specs are failing? Looks relevant

lib/temporal/workflow/poller.rb
robholland: If that were the case then we'd see "Unable to poll workflow task queue" all over the logs, but we don't. So I think failure here is unexpected.

antstorm: That's a good point, you're probably right. Please ignore it then

lib/temporal/workflow.rb
robholland: Done.

robholland:
Fixed the example specs failing, due to local workflow context not supporting metadata. I've added that now and adjusted callers to pass that rather than plain headers.

lib/temporal/workflow/poller.rb
robholland: This is required to catch things like an undefined workflow class. I don't really see a reason to treat this any differently. What's your concern here?

lib/temporal/activity/task_processor.rb
```diff
@@ -26,16 +29,17 @@ def process
...
 
-        result = middleware_chain.invoke(metadata) do
+        result = middleware_chain.invoke(metadata, task.to_h) do
           activity_class.execute_in_context(context, parse_payload(task.input))
         end
 
         # Do not complete asynchronous activities, these should be completed manually
         respond_completed(result) unless context.async?
       rescue StandardError, ScriptError => error
+        Temporal::ErrorHandler.handle(error, metadata: metadata.to_h)
```
antstorm, line 41: Btw, I see where you coming from with `.to_h`, but compared to logs it's actually that bad to pass the whole metadata object here. Any reasons you decided not to do it?

antstorm: Is would allow devs to distinguish between different metadata types

antstorm (changes requested):

lib/temporal/metadata/activity.rb
```diff
@@ -23,6 +23,21 @@ def initialize(namespace:, id:, name:, task_token:, attempt:, workflow_run_id:,
       def activity?
         true
       end
+
+      def to_h
+        {
+          'attempt' => attempt,
+          'headers' => headers,
```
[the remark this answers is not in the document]
robholland: Ok, I confess I don't really know what they are used for, I'll remove.

lib/temporal/metadata/activity.rb
```diff
@@ -23,6 +23,21 @@ def initialize(namespace:, id:, name:, task_token:, attempt:, workflow_run_id:,
...
+          'heartbeat_details' => heartbeat_details,
+          'activity_id' => id,
+          'activity_name' => name,
+          'namespace' => namespace,
+          'task_token' => task_token,
```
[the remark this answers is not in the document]
robholland: We used it to debug a Temporal/Ruby-SDK bug, not sure how useful it would be in general.

lib/temporal/metadata/activity.rb
```diff
@@ -23,6 +23,21 @@ def initialize(namespace:, id:, name:, task_token:, attempt:, workflow_run_id:,
       def activity?
         true
       end
+
+      def to_h
+        {
+          'attempt' => attempt,
+          'headers' => headers,
+          'heartbeat_details' => heartbeat_details,
```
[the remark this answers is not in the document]
robholland: This is important to send, the concept of heartbeat details is to provide cursors for long running activities. If you don't have the heartbeat details alongside an exception you can know which slice/chunk/batch it was working on when it blew up.

lib/temporal/activity/task_processor.rb
```diff
@@ -6,9 +6,12 @@
 module Temporal
   class Activity
     class TaskProcessor
+      attr_accessor :metadata
```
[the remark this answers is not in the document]
robholland: Right, doesn't need to be, will fix.

lib/temporal/activity/task_processor.rb
```diff
@@ -26,16 +29,17 @@ def process
           raise ActivityNotRegistered, 'Activity is not registered with this worker'
         end
 
-        metadata = Metadata.generate(Metadata::ACTIVITY_TYPE, task, namespace)
         context = Activity::Context.new(client, metadata)
 
-        result = middleware_chain.invoke(metadata) do
+        result = middleware_chain.invoke(metadata, task.to_h) do
```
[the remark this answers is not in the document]
robholland: If I had been able to see the task alongside the exception we saw then it would have been immediately obvious what the issue was (query being sent with no workflow tasks). Maybe I'm bias as an SDK contributor, but this information is extremely valuable to me.

lib/temporal/activity/task_processor.rb
robholland: I didn't want application developers to have to know the metadata class internals. If we send a hash then there is no API concerns. I don't feel strongly about it though if you'd prefer to send the objects I can `.to_h` in my application's handler.

antstorm:
@robholland can you please rebase it and I'll get it merged?

robholland:
Rebased.

antstorm (approved):

antstorm:
Thank you, @robholland , great PR!

PR #40, opened 2021-02-08
Draft of activity/second limiting.

robholland:


antstorm:
I'm assuming we're dropping/redoing this one in favour of using server-side rate limiting, right?

PR #42, opened 2021-02-17
Correct id selection for WorkflowExecutionTerminated event.

robholland:
Fixes: NoMethodError: undefined method `scheduled_event_id' for #<Temporal::Api::History::V1::WorkflowExecutionTerminatedEventAttributes:0x000056370c436d38>

### DIFF
--- a/lib/temporal/workflow/history/event.rb
+++ b/lib/temporal/workflow/history/event.rb
@@ -10,8 +10,6 @@ module Temporal
           ACTIVITY_TASK_CANCELED
           TIMER_FIRED
           REQUEST_CANCEL_EXTERNAL_WORKFLOW_EXECUTION_FAILED
-          WORKFLOW_EXECUTION_SIGNALED
-          WORKFLOW_EXECUTION_TERMINATED
           SIGNAL_EXTERNAL_WORKFLOW_EXECUTION_FAILED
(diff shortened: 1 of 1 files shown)


lib/temporal/workflow/history/event.rb
```diff
@@ -10,8 +10,6 @@ class Event
           ACTIVITY_TASK_CANCELED
           TIMER_FIRED
           REQUEST_CANCEL_EXTERNAL_WORKFLOW_EXECUTION_FAILED
-          WORKFLOW_EXECUTION_SIGNALED
-          WORKFLOW_EXECUTION_TERMINATED
```
antstorm, line 14: Why were these removed?

robholland: The constant is only used as part of this case statement and those two event types don't have the `scheduled_event_id` unlike the others in that list so they don't feel like they belong.

robholland: It's fine for them to stay if you'd prefer now terminated event is correctly "shadowed" by the proceeding `when`. Just not really sure what determines which events should be in the list.

antstorm: Ah, I see, good point 👍

antstorm (approved):

PR #43, opened 2021-02-17
Add more details to logs to improve debugging.

robholland:


robholland:
If you're happy with the approach I'll extend this to the workflow processor also.

robholland:
Sample from the examples:

```I, [2021-02-17T11:45:33.322764 #30755]  INFO -- temporal_client: Activity task completed (ruby-samples / HelloWorldWorkflow[852345e6-30e9-4860-8ad8-690d7f9f580a:025964be-877a-43a5-9ba2-06e647de9f55] / HelloWorldActivity[5]#1)```

**Review by tomharrisonjr — APPROVED** [review]
Nice additions!

File: lib/temporal/activity/task_processor.rb [hunk 1]
```diff
@@ -19,28 +19,29 @@ def initialize(task, namespace, activity_lookup, client, middleware_chain)
...
+        metadata = Metadata.generate(Metadata::ACTIVITY_TYPE, task, namespace)
+        context = Activity::Context.new(client, metadata)
+
         if !activity_class
           raise ActivityNotRegistered, 'Activity is not registered with this worker'
         end
 
-        metadata = Metadata.generate(Metadata::ACTIVITY_TYPE, task, namespace)
-        context = Activity::Context.new(client, metadata)
+        Temporal.logger.info("Activity task processing (#{context.log_tag})")
```
antstorm, line 31: this is probably "debug" level log

lib/temporal/activity/context.rb
```diff
@@ -56,6 +56,10 @@ def headers
         metadata.headers
       end
 
+      def log_tag
+        "#{metadata.namespace} / #{metadata.workflow_name}[#{metadata.workflow_id}:#{metadata.workflow_run_id}] / #{metadata.name}[#{metadata.id}]##{metadata.attempt}"
```
antstorm, line 60: this can be generated once, no need to generate it every time the method is called

lib/temporal/activity/context.rb
```diff
@@ -56,6 +56,10 @@ def headers
         metadata.headers
       end
 
+      def log_tag
```
antstorm, line 59: This method definitely belongs on `Metadata` and not `Context`, this is quite clear from the fact that it only calls "metadata" in it

antstorm (changes requested):
I like the general direction of adding more info, but let's do this slightly differently. Unstructured logging is difficult to parse and work with. Instead let's add an optional hash to the log calls with all the additional information we want to pass.

Many companies are already doing structured logging, so when implementing their own adapter everyone should be able to format the log lines as they prefer

lib/temporal/activity/context.rb
robholland: This method is removed now.

lib/temporal/activity/context.rb
robholland: Method removed.

lib/temporal/activity/task_processor.rb
```diff
@@ -21,29 +21,30 @@ def initialize(task, namespace, activity_lookup, client, middleware_chain)
       def process
         start_time = Time.now
 
-        Temporal.logger.info("Processing activity task for #{activity_name}")
+        Temporal.logger.debug("Processing Activity task", metadata.to_h)
         Temporal.metrics.timing('activity_task.queue_time', queue_time_ms, activity: activity_name)
 
+        metadata = Metadata.generate(Metadata::ACTIVITY_TYPE, task, namespace)
```
antstorm, line 27: This is already done on line 13, right?

lib/temporal/activity/task_processor.rb
```diff
@@ -56,20 +57,20 @@ def queue_time_ms
         ((started - scheduled) * 1_000).round
       end
 
-      def respond_completed(result)
-        Temporal.logger.info("Activity #{activity_name} completed")
+      def respond_completed(result, context)
```
antstorm, line 60: I'm assuming the intention was to pass `metadata`, not `context`. You also don't need to do this since metadata is available via the `attr_reader`

lib/temporal/activity/task_processor.rb
```diff
@@ -56,20 +57,20 @@ def queue_time_ms
...
       rescue StandardError => error
-        Temporal.logger.error("Unable to complete Activity #{activity_name}: #{error.inspect}")
+        Temporal.logger.error("Unable to complete Activity", metadata.to_h.merge(error: error.inspect))
 
         Temporal::ErrorHandler.handle(error, metadata: metadata)
       end
 
-      def respond_failed(error)
-        Temporal.logger.error("Activity #{activity_name} failed with: #{error.inspect}")
+      def respond_failed(error, context)
```
antstorm, line 69: same here

lib/temporal/workflow/task_processor.rb
```diff
@@ -40,16 +40,13 @@ def process
...
         complete_task(commands)
       rescue StandardError => error
-        fail_task(error)
-
-        Temporal.logger.error("Workflow task for #{workflow_name} failed with: #{error.inspect}")
-        Temporal.logger.debug(error.backtrace.join("\n"))
-
         Temporal::ErrorHandler.handle(error, metadata: metadata)
+
+        fail_task(error)
```
antstorm, line 45: This is a change in behaviour — failing a task after handling errors, can you please elaborate?

antstorm (changes requested):
Found a couple of potential issues. Also can you please add a spec for the new logger class?

lib/temporal/activity/task_processor.rb
robholland: Oops, left over from a previous revision.

lib/temporal/workflow/task_processor.rb
robholland: That's how the workflow one does it. I don't have a preference, but I think it should be consistent.

lib/temporal/workflow/replay_aware_logger.rb
```diff
@@ -11,17 +11,17 @@ def initialize(main_logger, replay = true)
       end
 
       SEVERITIES.each do |severity|
-        define_method severity do |message|
+        define_method severity do |*args|
```
antstorm, line 14: Can we keep this interface same as our main logger? E.g.:

```ruby
define_method severity do |message, details = {}|
  ...
end
```

lib/temporal/workflow/replay_aware_logger.rb
```diff
@@ -11,17 +11,17 @@ def initialize(main_logger, replay = true)
...
           return if replay?
 
-          main_logger.public_send(severity, message)
+          main_logger.public_send(severity, *args)
         end
       end
 
-      def log(severity, message)
+      def log(severity, *args)
```
antstorm, line 21: same here

lib/temporal/workflow/task_processor.rb
antstorm: Got it, makes sense, didn't notice these were different to begin with

lib/temporal/workflow/replay_aware_logger.rb
robholland: And this.

tristan-instacart:
This change meant that a standard logger passed into config will no longer function and all loggers in effect must be of type Temporal::Logger because of the signature change on the standard log methods. 

This makes things a little awkward if you have standard log practices in your organization. It is probably worth a call out in the README that while logger is exposed in Temporal.configure, you should use caution overriding it.

antstorm:
@tristan-instacart you're right, apologies for this. Hopefully the new interface is very easily adoptable.

@robholland would you mind updating the README with the mention of the new interface?

PR #44, opened 2021-02-19
Add to_h for metadata.

robholland:
This will be used in logging/error reporting.

### DIFF
--- a/lib/temporal/metadata/activity.rb
+++ b/lib/temporal/metadata/activity.rb
@@ -23,6 +23,21 @@ module Temporal
       def activity?
         true
       end
+
+      def to_h
+        {
+          'attempt' => attempt,
+          'headers' => headers,
+          'heartbeat_details' => heartbeat_details,
+          'activity_id' => id,
+          'activity_name' => name,
+          'namespace' => namespace,
+          'task_token' => task_token,
+          'workflow_id' => workflow_id,
+          'workflow_name' => workflow_name,
+          'workflow_run_id' => workflow_run_id
+        }
+      end
     end
   end
 end
--- a/lib/temporal/metadata/activity.rb
+++ b/lib/temporal/metadata/activity.rb
@@ -26,16 +26,14 @@ module Temporal
 
       def to_h
         {
-          'attempt' => attempt,
-          'headers' => headers,
-          'heartbeat_details' => heartbeat_details,
-          'activity_id' => id,
-          'activity_name' => name,
           'namespace' => namespace,
-          'task_token' => task_token,
           'workflow_id' => workflow_id,
           'workflow_name' => workflow_name,
-          'workflow_run_id' => workflow_run_id
+          'workflow_run_id' => workflow_run_id,
+          'activity_id' => id,
+          'activity_name' => name,
+          'attempt' => attempt,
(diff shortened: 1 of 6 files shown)


lib/temporal/metadata/activity.rb
```diff
@@ -23,6 +23,21 @@ def initialize(namespace:, id:, name:, task_token:, attempt:, workflow_run_id:,
       def activity?
         true
       end
+
+      def to_h
+        {
+          'attempt' => attempt,
+          'headers' => headers,
```
antstorm, line 30: Commented on the other PR, replaying here: headers might not be safe to pass to logging (in case sensitive info is passed), so we might want to exclude it

lib/temporal/metadata/activity.rb
```diff
@@ -23,6 +23,21 @@ def initialize(namespace:, id:, name:, task_token:, attempt:, workflow_run_id:,
...
+          'heartbeat_details' => heartbeat_details,
```
antstorm, line 31: Same here ^

lib/temporal/metadata/activity.rb
```diff
@@ -23,6 +23,21 @@ def initialize(namespace:, id:, name:, task_token:, attempt:, workflow_run_id:,
...
+          'activity_id' => id,
+          'activity_name' => name,
+          'namespace' => namespace,
+          'task_token' => task_token,
```
antstorm, line 35: I don't think this is useful to anyone

antstorm:
Btw, thanks for splitting this one into a separate PR 👍

lib/temporal/metadata/activity.rb
robholland: We actually just used this as part of the information required to debug the cron workflow issue we were discussing on slack. We noticed the task token was a query token, not a workflow task token. Granted I'm not sure how useful it would normally be.

robholland:
Now that I think about it, shouldn't it be for developers to decide what they want to send on to their error reporting or logging tool? I agree it would be possible to leak information from headers, but the developer chooses which fields they relay. Also using something like Sentry for example, it would recognise and filter out the value for Authorization headers using it's default "scrubbers".

robholland:
Discussed this on slack. Error/log APIs will send metadata objects not the hash, developers can add anything they like into the hash before logging/reporting if they need.

lib/temporal/metadata/activity.rb
```diff
@@ -23,6 +23,19 @@ def initialize(namespace:, id:, name:, task_token:, attempt:, workflow_run_id:,
...
+      def to_h
+        {
+          'namespace' => namespace,
+          'workflow_id' => workflow_id,
+          'workflow_name' => workflow_name,
+          'workflow_run_id' => workflow_run_id,
+          'activity_id' => id,
+          'activity_name' => name,
+          'attempt' => attempt,
+          'heartbeat_details' => heartbeat_details
```
anthonydmitriyev, line 36: If we're leaving `heartbeat_details`, should we be concerned with decoding the information in them? You'd expect plain values in this hash, but `heartbeat_details` does sound like it is it's own world

antstorm (approved):
👍

PR #46, opened 2021-02-25
Allow varying options per activity task queue.

robholland:
This also adds support for server-side rate limiting of activities per
second for a given task queue.

### DIFF
--- a/lib/temporal/worker.rb
+++ b/lib/temporal/worker.rb
@@ -16,9 +16,10 @@ module Temporal
       @workflow_task_middleware = []
       @activity_middleware = []
       @shutting_down = false
-      @activity_poller_options = {
+      @default_activity_poller_options = {
         thread_pool_size: activity_thread_pool_size,
       }
+      @activity_poller_options = Hash.new { {} }
     end
 
     def register_workflow(workflow_class, options = {})
@@ -33,6 +34,10 @@ module Temporal
       key = [execution_options.namespace, execution_options.task_queue]
 
       @activities[key].add(execution_options.name, activity_class)
+    end
+
+    def configure_activity_task_queue(name, options = {})
+      @activity_poller_options[name] = @activity_poller_options[name].merge(options)
     end
 
     def add_workflow_task_middleware(middleware_class, *args)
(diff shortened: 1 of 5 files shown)


lib/temporal/worker.rb
```diff
@@ -9,7 +9,7 @@ module Temporal
   class Worker
     # activity_thread_pool_size: number of threads that the poller can use to run activities.
     #   can be set to 1 if you want no paralellism in your activities, at the cost of throughput.
-    def initialize(activity_thread_pool_size: Temporal::Activity::Poller::DEFAULT_OPTIONS[:thread_pool_size])
+    def initialize(activity_thread_pool_size: Temporal::Activity::Poller::DEFAULT_OPTIONS[:thread_pool_size], activity_max_tasks_per_second: nil)
```
jeffschoner-stripe, line 12: Should this be a hash of options? There are [quite a few options in the Go SDK](https://pkg.go.dev/go.temporal.io/sdk@v1.5.0/internal#WorkerOptions) for worker configuration that may be eventually added here. Changing this later would be a breaking change.

Both [Java](https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/worker/WorkerOptions.html) and Go SDKs name this "max task queue activities per second". Admittedly that does not match the field name in the proto. Because there is a distinction made between this limit and "max worker activities per second" too, it might make sense to add both options with explanatory comments while you're already in here.

robholland: Changing this to an options hash later would not break the API. That would be backward compatible with the current kwargs. Unless you mean doing `initialize(options: { ... })` but I can't see what the value is in doing that? Happy to change the option name to be closer to the Go SDK one. I don't think we should add the other option as part of this PR though, it would require adding code to do the rate limiting that isn't related to the server side rate limiting.

spec/unit/lib/temporal/worker_spec.rb
```diff
@@ -180,7 +180,23 @@ class TestWorkerActivity < Temporal::Activity
...
+        .and_return(activity_poller)
+
+      worker = Temporal::Worker.new(activity_max_tasks_per_second: 15)
+      allow(worker).to receive(:shutting_down?).and_return(true)
+      worker.register_workflow(TestWorkerWorkflow)
+      worker.register_activity(TestWorkerActivity)
+
+      worker.start
+
+      expect(activity_poller).to have_received(:start)
```
antstorm, line 199: nit: I'm not sure this check is needed in this spec, we're only checking that the Poller is initialised with the new option

PR #47, opened 2021-03-01
UTF safety

robholland:
Fixes Encoding::UndefinedConversionError errors for some UTF characters.

### DIFF
--- a/lib/temporal/client/serializer/payload.rb
+++ b/lib/temporal/client/serializer/payload.rb
@@ -21,7 +21,7 @@ module Temporal
             payloads: [
               Temporal::Api::Common::V1::Payload.new(
                 metadata: { 'encoding' => JSON_ENCODING },
-                data: JSON.serialize(object)
+                data: JSON.serialize(object).b
               )
             ]
           )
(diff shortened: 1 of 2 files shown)


antstorm:
@robholland can you help me understand where this error comes from? Is it protobuf encoding complaining or GRCP? Or Temporal server?

robholland:
It's the transcoding attempt that the protobuf code makes to try and turn the serialised JSON string from us (which contains UTF-8 characters which are valid in JSON) into ASCII-8BIT to fit in the `bytes`-typed data field for Temporal Payloads. UTF characters with the high bit set are not transcodable into ASCII-8BIT as they have no ascii equivalent so the transcode fails. Calling `.b` on the serialised JSON string escapes any high-bit-set characters which allows the string to be transcoded to ASCI-8BIT safely, and then deserialised on the other side into the correct UTF-8 characters.

robholland:
Prior art here: https://github.com/protocolbuffers/protobuf/issues/4228

antstorm (approved):
👍

PR #49, opened 2021-03-12
Support workflow termination.

robholland:


lib/temporal/client/grpc_client.rb
```diff
@@ -273,16 +273,32 @@ def reset_workflow_execution(namespace:, workflow_id:, run_id:, reason:, workflo
           workflow_execution: Temporal::Api::Common::V1::WorkflowExecution.new(
             workflow_id: workflow_id,
             run_id: run_id,
-            request_id: SecureRandom.uuid
```
robholland, line 276: This isn't a field on `WorkflowExecution` but on the higher `ResetWorkflowExecutionRequest` so I don't think this method would ever have worked.

robholland: Also it's optional, so providing a random string that we don't return anyway doesn't feel useful.

lib/temporal.rb
```diff
@@ -36,6 +36,18 @@ def start_workflow(workflow, *input, **args)
       response.run_id
     end
 
+    def terminate_workflow(workflow, workflow_id:, **options)
```
antstorm, line 39: Can you please make it look like https://github.com/coinbase/cadence-ruby/pull/19?

lib/temporal.rb
```diff
@@ -36,6 +36,18 @@ def start_workflow(workflow, *input, **args)
...
+      execution_options = ExecutionOptions.new(workflow, options)
+
+      client.terminate_workflow_execution(
+        namespace: execution_options.namespace,
+        workflow_id: workflow_id,
+        run_id: options[:run_id],
```
antstorm, line 45: Not sure if `run_id` is actually optional here and what would happen if there are multiple executions?

antstorm (changes requested):

lib/temporal.rb
robholland: `run_id` is optional: https://github.com/temporalio/sdk-go/blob/27e8be2e856181ef2b97e34690f6ab085d3170ce/client/client.go#L162

lib/temporal.rb
robholland: Do you mean the method signature? I think it should line up with start_workflow/schedule_workflow, and they use kwargs. `terminate_workflow_execution` takes namespace as an argument so I think ExecutionOptions should be used so that the Workflow classes namespace configuration is taken into account (as for start/schedule).

robholland: Updated the method signature, although left `namespace` as nillable because it has a default value we can use.

antstorm (approved):

PR #53, opened 2021-03-18
Support multiple payloads from array input

gvieira:
We have a workflow written in Go that receives two args.

```go
func Workflow(ctx workflow.Context, a string, b string) error {
```

When calling it from a different go program, everything works as expected.

```go
we, err := c.ExecuteWorkflow(context.Background(), workflowOptions, "Workflow", "a.json", "b.json")
```

This is a simple screenshot of the input, from `localhost:8088`.

Screen Shot 2021-03-18 at 20 45 10

However, when calling the very same workflow from Ruby, the current code doesn't work.

```ruby
Temporal.start_workflow('Workflow', 'a.json', 'b.json', options: options)
```

Here's a screenshot of the input in this case:

Screen Shot 2021-03-18 at 20 45 27

We get the following error:

```
wrapError: unable to decode the workflow function input payload with error: payload item 0: unable to decode: json: cannot unmarshal array into Go value of type string
```

... which makes a lot of sense, as it tries to put an array into a string.

-----------

If you dive in and see the code in Go that encodes payloads, you'll end up with something like this:

```go
dataConverter := converter.GetDefaultDataConverter()
input, _ := dataConverter.ToPayloads("a.json", "b.json")
fmt.Println(input.Payloads[0])
fmt.Println(input.Payloads[1])
```

Here, `input` ends up with 2 payloads.

But if you try to do the very same thing in Ruby, you'll get something like this:

```ruby
Temporal::Client::Serializer::Payload.new(['a.json', 'b.json']).to_proto
```

In this case, with the current code, you are going to get a single payload, which encodes the array, hence the wrong input you can see above.

-------

All I've done is to map the array into payloads. I monkey-patched it into the application (for testing purposes) and it worked.

### DIFF
--- a/lib/temporal/activity/task_processor.rb
+++ b/lib/temporal/activity/task_processor.rb
@@ -31,7 +31,7 @@ module Temporal
         context = Activity::Context.new(client, metadata)
 
         result = middleware_chain.invoke(metadata) do
-          activity_class.execute_in_context(context, parse_payload(task.input))
+          activity_class.execute_in_context(context, Client::Serializer::Payload.from_proto(task.input))
         end
 
         # Do not complete asynchronous activities, these should be completed manually
--- a/lib/temporal/client/serializer/payload.rb
+++ b/lib/temporal/client/serializer/payload.rb
@@ -10,20 +10,34 @@ module Temporal
         def self.from_proto(proto)
           return if proto.nil? || proto.payloads.empty?
 
-          binary = proto.payloads.first.data
-          JSON.deserialize(binary)
+          data = proto.payloads.map do |payload|
+            JSON.deserialize(payload.data)
+          end
+
+          data.size == 1 ? data.first : data
         end
 
         def to_proto
           return if object.nil?
 
+          payloads = case object
+                     when Array
+                       object.map(&method(:payload_for))
+                     else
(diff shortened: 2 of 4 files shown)


antstorm (commented):
@gvieira thank you for contributing! The code looks good, but can you please explain the reasoning behind this change? Array will already get correctly converted to and from a proto.

A spec exposing this would be useful (the one you've added checks that it results in multiple payloads, but converting it back results in the same array as originally submitted)

gvieira:
Hey @antstorm! Thanks for taking a look at this. I was planning to go through this with a teammate before moving forward and asking for review, but I forgot to set as draft.

Anyway, I've added a description here. I was hoping that [this](https://github.com/coinbase/temporal-ruby/pull/53/files#diff-08901e0e9a9cf55ba6bc099069fd9b962ccab32b5ee707b75c51345cef1e684aR37) spec would help with the understanding of what I was trying to fix.

robholland:
I've added an alternative solution to this which brings the Ruby SDK payload handling in line with the other SDKs and avoids edge cases: https://github.com/coinbase/temporal-ruby/pull/56

gvieira:
Closing in favor of #56.

PR #55, opened 2021-03-18
Integration specs against a running server

antstorm:
This PR adds support for running specs in CircleCI against a running Temporal server. This allows for implementing true end-to-end specs observing the result and history of the execution

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,7 @@ jobs:
           - POSTGRES_USER=postgres
           - POSTGRES_PWD=temporal
           - POSTGRES_SEEDS=postgres
+          - TEMPORAL_HOST=temporal
           - DYNAMIC_CONFIG_FILE_PATH=config/dynamicconfig/development.yaml
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,8 +29,11 @@ jobs:
           - POSTGRES_USER=postgres
           - POSTGRES_PWD=temporal
           - POSTGRES_SEEDS=postgres
-          - TEMPORAL_HOST=temporal
           - DYNAMIC_CONFIG_FILE_PATH=config/dynamicconfig/development.yaml
+
+    environment:
+      - TEMPORAL_HOST=temporal
+
     steps:
       - checkout
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,11 +17,35 @@ jobs:
   test_examples:
     docker:
       - image: circleci/ruby:2.6.3-stretch-node
+      - image: circleci/postgres:alpine
+        name: postgres
+        environment:
+          POSTGRES_PASSWORD: temporal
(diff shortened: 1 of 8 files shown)


examples/spec/integration/serial_hello_world_workflow_spec.rb
```diff
@@ -3,6 +3,37 @@
...
+    )
+
+    client = Temporal.send(:client)
+
+    result = client.get_workflow_execution_history(
+      namespace: Temporal.configuration.namespace,
+      workflow_id: workflow_id,
+      run_id: run_id,
+      next_page_token: nil,
+      wait_for_new_event: true,
```
drewhoskins-stripe, line 30: I imagine even with `wait_for_new_event` you
(a) need to run this command first to avoid the case where the workflow finishes first.  (unless `wait_new_event` returns immediately if the the workflow has already hit a terminal state?)
(b) still need a polling loop until it gets to a terminal state in case there are intermediary events.

examples/spec/integration/serial_hello_world_workflow_spec.rb
```diff
@@ -3,6 +3,37 @@
 describe SerialHelloWorldWorkflow do
   subject { described_class }
 
+  it 'works' do
+    Temporal.configure do |config|
```
drewhoskins-stripe, line 7: We can factor this config block into a shared helper.

drewhoskins-stripe:
Overall quite exciting!  Some documentation as you polish this would be great!

examples/spec/integration/serial_hello_world_workflow_spec.rb
antstorm: Great points:

(a) I don't think you can run this first since you won't have a run_id at that time. I'll test it, but I'm pretty sure it'll return new events since the last call
(b) `event_type: :close` will filter out any intermediary events

.circleci/config.yml
```diff
@@ -17,11 +17,35 @@ jobs:
...
+      - image: temporalio/auto-setup:latest
+        name: temporal
+        environment:
+          - DB=postgresql
+          - DB_PORT=5432
+          - POSTGRES_USER=postgres
+          - POSTGRES_PWD=temporal
+          - POSTGRES_SEEDS=postgres
+          - DYNAMIC_CONFIG_FILE_PATH=config/dynamicconfig/development.yaml
     steps:
```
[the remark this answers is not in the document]
antstorm: This are CircleCI specific configuration, but you should be able to just run the integration spec with the Temporal running in docker along with the worker and get the same results

examples/spec/integration/serial_hello_world_workflow_spec.rb
```diff
@@ -3,6 +3,37 @@
 describe SerialHelloWorldWorkflow do
   subject { described_class }
 
+  it 'works' do
```
robholland, line 6: `completes`?

robholland (approved):

PR #56, opened 2021-03-19
Refactor serialization to converter.

robholland:
This follows the patterns of the Go and Java SDKs and allows
interoperability between Ruby and the other SDKs by converting arguments
to/from payloads in the same way.

Serialisation of protobufs has been left as-is because it is not
affected by interoperability issues.


antstorm (changes requested):
Overall great change and I really like the direction. Left a few inline comments.

Besides that:
1. Let's make it backwards compatible by adding a catch-all `Legacy` converter that doesn't has a `to_payload` method, only `from_payload`. People can add it to the stack and remove it later after they have migrated all the workflows
2. We might wanna leave the `Serializer` that uses converters instead of the Composite setup that makes little sense in this case (normally you'd use Composite with "mixed" encoding in case you need to nest these recursively, but there's probably no reason in doing so). This should also reduce the diff leaving the `Serializer` calls in place

lib/temporal/client/converter/composite.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        def initialize(converters:)
+          @converters = converters
+          @converters_by_encoding = {}
+          converters.each do |converter|
+            @converters_by_encoding[converter.encoding] = converter
+          end
+        end
+
+        def from_payload(payload)
```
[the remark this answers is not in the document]
robholland: It isn't, I just hadn't filled out the API. I've refactored to have a Base converter, hopefully that makes intent more obvious. Not all protobufs have a `payloads`, for example `details` is usually a single `payload`. The API needs to support directly wrapping individual payloads as well.

lib/temporal/client.rb
```diff
@@ -17,5 +21,19 @@ def self.generate
 
       client_class.new(host, port, identity)
     end
+
+    def self.default_converter
```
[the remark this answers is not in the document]
robholland: I explained in slack the reason for this is to pave the way for the refactor you did to support multiple clients. I can move this to configuration for now though if you prefer.

robholland:
I don't see how we would know which `payload`s are legacy and which are not, they both set encoding as `json/plain` but one has a nested array which should be unpacked and one does not. Testing for an array feels very brittle.

robholland:
The composite encoder is to allow people to match the encoding stacks they are using from other SDKs. It's the same system used in Go/Java SDKs.

spec/unit/lib/temporal/client/converter_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+      it 'decodes an null payload to nil' do
+        payload = Temporal::Api::Common::V1::Payload.new(
+          metadata: { 'encoding' => 'binary/null' }
+        )
+
+        expect(subject.from_payload(payload)).to eq(nil)
+      end
+    end
+
+    describe 'for a bytestring payload' do
```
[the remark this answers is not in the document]
robholland: Split out.

spec/unit/lib/temporal/client/converter_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+        input2 = [2]
+
+        expect(subject.from_payloads(subject.to_payloads(input1, input2))).to eq([input1, input2])
+      end
+    end
+
+    describe 'for nil payload' do
+      it 'encodes to a null payload' do
+        payload = Temporal::Api::Common::V1::Payload.new(
+          metadata: { 'encoding' => 'binary/null' }
```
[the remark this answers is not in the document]
robholland: Personally I think that makes the spec less readable but I've made this change.

lib/temporal/client/converter/composite.rb
```diff
@@ -0,0 +1,57 @@
+module Temporal
+  module Client
+    module Converter
+      class Composite
```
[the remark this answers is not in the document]
robholland: Composite now presents the same API as the others.

robholland:
Added a Legacy converter as discussed on Slack. The idea would be to roll the gem out with `Temporal.configure { |config| config.converter = Temporal::Client::Converters::Legacy.new }`. Once all workflows have completed or been restarted all payloads should be marked with `json/legacy`. At this point it should be safe to remove the converter configuration, the default stack will encode new payloads with the correct semantics but be able to handle legacy payloads when reading.

spec/fabricators/grpc/activity_task_fabricator.rb
```diff
@@ -6,14 +6,14 @@
   activity_id { SecureRandom.uuid }
   task_token { |attrs| attrs[:task_token] || SecureRandom.uuid }
   activity_type { Fabricate(:api_activity_type) }
-  input { Temporal::Api::Common::V1::Payloads.new(payloads: [Fabricate(:api_payload)]) }
+  input { Temporal.configuration.converter.to_payloads() }
```
antstorm, line 9: nit: no need for `()`

antstorm (commented):
@robholland can you help me understand your thinking behind the base class for converters? More specifically — the composite converter stands out from others by quite a lot and makes a bit more sense as it's own separate object and not a subclass of a converter base

You are also missing specs for the composite converter

robholland:
The composite converter provides the same API as the others, they are interchangable.

lib/temporal/client/converter/composite.rb
```diff
@@ -0,0 +1,57 @@
...
+        class ConverterNotFound < Exception; end
```
Scamboy23: opus-grove:rh-payload-handling

lib/temporal/client/converter/bytes.rb
```diff
@@ -0,0 +1,28 @@
...
+      class Bytes
+        ENCODING = 'binary/plain'.freeze
+
+        def encoding
+          ENCODING
+        end
+
+        def from_payload(payload)
+          payload.data
+        end
```
Scamboy23, line 16: opus-grove:rh-payload-handling

lib/temporal/client/converter/nil.rb
```diff
@@ -0,0 +1,25 @@
+module Temporal
+  module Client
+    module Converter
```
Scamboy23, line 5: lib/temporal/client/converter/nil.rb

Scamboy23 (commented):
lib/temporal/client/converter/nil.rb

lib/temporal/client/converter/bytes.rb
Scamboy23: https://github.com/RubyMoney/money-heuristics/blob/13c5e214fae697b556758d6a7cf2a65de464d6ad/lib/money-heuristics/analyzer.rb

antstorm:
@robholland I think you're mixing 2 separate interfaces together making them broader than they should be. This is how I see it — composite interface deals with `Payloads`, while all other converters deal with a `Payload` with a known "encoding". This way whoever is using the gem can easily extend it:

```
configuration.add_converter_for("my_encoding", MyConverter)
configuration.add_converter_for("my_encoding_2", MyConverter2)
```

without having to redefine the whole composite. It is also much clearer which converters are being override this way (if any)

antstorm:
@robholland I actually think that `Base` class for converters was a good idea. If anything it'll demonstrate how people can implement their own converters. The problem was that `Composite` does not fit that interface — it doesn't have `encoding` and has extra methods.

Please let me know if you want to get on a call to resolve this, otherwise I feel like we're getting stuck here

antstorm:
> The composite converter provides the same API as the others, they are interchangable.

Clearly they are not because you can't do:

```ruby
Converter::Composite.new(converters: [Converter::Composite.new])
```

robholland:
Please can you just make whichever changes you would like to get this merged?

examples/bin/worker
```diff
@@ -14,6 +14,7 @@ worker.register_workflow(AsyncHelloWorldWorkflow)
 worker.register_workflow(BranchingWorkflow)
 worker.register_workflow(CancellingTimerWorkflow)
 worker.register_workflow(CheckWorkflow)
+worker.register_workflow(ComplexResultsWorkflow)
```
antstorm, line 17: @robholland I think you're missing these new workflows and activities from the PR

robholland: Oops! Added now.

antstorm:
@robholland another thing I've noticed — we are not yet using new converters for converting activity or workflow headers. Is this intentional or just accidentally missing?

robholland:
It was missed, the current code doesn't serialize them, I don't know if the current code for headers actually works?

antstorm:
@robholland it should work, yes. It's also pretty straightforward, I believe each header is serialized using a single payload

lib/temporal/client/grpc_client.rb
```diff
@@ -206,7 +204,7 @@ def respond_activity_task_completed_by_id(namespace:, activity_id:, workflow_id:
           workflow_id: workflow_id,
           run_id: run_id,
           activity_id: activity_id,
-          result: Serializer::Payload.new(result).to_proto
+          result: Temporal.configuration.converter.to_payloads(result)
```
antstorm, line 207: Is this intended not to wrap the result in array like the `respond_activity_task_completed` call?

lib/temporal/client/grpc_client.rb
```diff
@@ -235,7 +233,7 @@ def respond_activity_task_failed_by_id(namespace:, activity_id:, workflow_id:, r
       def respond_activity_task_canceled(task_token:, details: nil)
         request = Temporal::Api::WorkflowService::V1::RespondActivityTaskCanceledRequest.new(
           task_token: task_token,
-          details: Serializer::Payload.new(details).to_proto,
+          details: Temporal.configuration.converter.to_payloads(details),
```
antstorm, line 236: Also this ^

lib/temporal/client/serializer/complete_workflow.rb
```diff
@@ -10,7 +9,7 @@ def to_proto
             command_type: Temporal::Api::Enums::V1::CommandType::COMMAND_TYPE_COMPLETE_WORKFLOW_EXECUTION,
             complete_workflow_execution_command_attributes:
               Temporal::Api::Decision::V1::CompleteWorkflowExecutionCommandAttributes.new(
-                result: Payload.new(object.result).to_proto
+                result: object.result.nil? ? nil : Temporal.configuration.converter.to_payloads([object.result])
```
antstorm, line 12: I wonder why we need this exception? If we always expect the payload data to be an array we can probably avoid special-casing `nil` results. Or am I missing something?

antstorm (commented):
@robholland so the plan is to always encode all elements of a top-level array as individual `Payload`s, right? It seems a bit weird when you get nested arrays and only top-level one gets special treatment, others just appear as JSON:

```ruby
>> Temporal.configuration.converter.to_payloads([nil, 1, 2, { a: :b }, [nil], [1, 2]])
=> <Temporal::Api::Common::V1::Payloads: payloads: [
     <Temporal::Api::Common::V1::Payload: metadata: {"encoding"=>"binary/null"}, data: "">, 
     <Temporal::Api::Common::V1::Payload: metadata: {"encoding"=>"json/plain"}, data: "1">, 
     <Temporal::Api::Common::V1::Payload: metadata: {"encoding"=>"json/plain"}, data: "2">, 
     <Temporal::Api::Common::V1::Payload: metadata: {"encoding"=>"json/plain"}, data: "{\":a\":\":b\"}">, 
     <Temporal::Api::Common::V1::Payload: metadata: {"encoding"=>"json/plain"}, data: "[null]">, 
     <Temporal::Api::Common::V1::Payload: metadata: {"encoding"=>"json/plain"}, data: "[1,2]">
   ]>
```

Just wondering how this is handled in other SDKs

lib/temporal/client/serializer/complete_workflow.rb
robholland: `Payloads` with one `Payload` which is `nil`, is not the same as having `Payloads` itself as `nil` which is what this does. This represents the difference between `return []` and `return nil` from a workflow.

robholland:
The Go and Java SDKs would produce the same `Payloads` structure given those workflow inputs. As a caller of `to_payloads` directly you see nested arrays, but note that a user does not. They would just pass `inputs` as an option to `start_workflow` or similar. `inputs` becomes the outer array and is special cased because we need to differentiate between individual `inputs` so that they go into their own payload. The Java and Go SDKs would produce the same payloads structure when given those same inputs (translated to their appropriate primitive types).

lib/temporal/client/grpc_client.rb
robholland: Nope, that's a bug.

lib/temporal/client/grpc_client.rb
robholland: Also a bug.

etherops:
Hi @robholland and @antstorm, thanks so much for putting this together guys!  We would really love to use this over at Instacart where we're deploying temporal.  We've currently implemented e2e workflow/activity encryption using the standard SDK DataConverter configurations.  We were going to work around the lack of DataConverter in the coinbase ruby SDK but saw this issue!  Is work on this paused or will this resume?  We can wait a bit on delivery if it's near completion and can even help contribute if needed.  Thanks!

robholland:
Hey @etherops, this is still active. I'll be working on creating an example that's interoperable with the cryptconverter in https://github.com/temporalio/samples-go/tree/f9c899850c39338df85df156a958ceb9e035cc5a/cryptconverter

etherops:
@robholland this is fantastic, thank you! is there anything I can do to help get the feature merged?

robholland:
@etherops testing alongside the go code is great if you have the opportunity, but we plan to get this reviewed, polished and merged soon.

examples/lib/cryptconverter.rb
```diff
@@ -0,0 +1,95 @@
...
+        end
+
+        def to_payloads(data)
+          key_id = get_key_id
+          key = get_key(key_id)
+
+          payloads = super(data)
+
+          Temporal::Api::Common::V1::Payloads.new(
+            payloads: payloads.payloads.map { |payload| encrypt_payload(payload, key_id, key) }
```
antstorm, line 30: So with this approach the mime type stays the same even though the payload gets replaced with a crypt, right? Is this how other SDKs do this as well? I'd expect something about the encoding to tell if this has been encrypted or not, like "application/json+encrypted" or something. Otherwise not using the right converter will result in really odd errors

lib/temporal/client/converter/base.rb
```diff
@@ -0,0 +1,51 @@
...
+        end
+
+        def to_payloads(data)
+          return nil if data.nil?
+          Temporal::Api::Common::V1::Payloads.new(
+            payloads: data.map(&method(:to_payload))
+          )
+        end
+
+        def from_result_payloads(payloads)
```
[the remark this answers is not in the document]
robholland: I can invert it but I worry it will lead to more bugs because then the caller always needs to perform this itself for results, details etc.

examples/lib/cryptconverter.rb
robholland: The metadata of the encrypted payload is 'binary/encrypted' https://github.com/coinbase/temporal-ruby/pull/56/files#diff-c6da22fe11b91519511828bd740132ec6b7aa0c54d50bd2a03c65242782a5b01R62. It doesn't matter what the content type is inside because the whole inner payload (metadata + data) is serialised + encrypted. After decryption we have the original payload back (with the original content type in metadata, `application/json` or whatever) which is then passed to the normal composite converter.

examples/bin/crypt_worker
```diff
@@ -0,0 +1,27 @@
...
+
+require 'temporal/worker'
+
+Dir[File.expand_path('../workflows/*.rb', __dir__)].each { |f| require f }
+Dir[File.expand_path('../activities/*.rb', __dir__)].each { |f| require f }
+Dir[File.expand_path('../middleware/*.rb', __dir__)].each { |f| require f }
+
+Temporal.configure do |config|
+  config.task_queue = 'crypt'
+  config.converter = Temporal::Client::Converter::Crypt.new(
```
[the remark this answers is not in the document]
robholland: This isn't a change, the crypt converter in both java and go wrap the default converters. This isn't something where I can just add a payload converter to the list of the composite converter because here we need to nest calls. The crypt converter runs the default converter and then wraps those payloads when encrypting, and vice versa when decrypting.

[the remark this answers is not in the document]
robholland: Go version: https://github.com/temporalio/samples-go/blob/master/cryptconverter/worker/main.go#L19

lib/temporal/client/converter/base.rb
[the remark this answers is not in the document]
robholland: Maybe an abstraction layer to handle this?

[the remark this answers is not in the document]
robholland: Will move these to a helper as discussed on the call.

lib/temporal/client/serializer/continue_as_new.rb
```diff
@@ -1,6 +1,4 @@
-require 'temporal/client/serializer/base'
-require 'temporal/client/serializer/payload'
```
antstorm, line 2: This file is still using `Payload` serializer, which is no longer there. Needs to be replaced and probably needs a spec to confirm

antstorm (changes requested):

antstorm (approved):
This is great, thank you so much for all the effort, @robholland!

PR #57, opened 2021-03-26
`require 'set'` whenever `Set` is used.

viettrungluu:
* You need to `require 'set'` before using `Set`.
* This is actually needed, but you rarely run into it since usually *someone*
  has already done it.
* You'd run into this with extremely minimal examples though (I ran into it
  with a minimal worker, in `Workflow::StateManager`).


antstorm (approved):
Great catch, thank you @viettrungluu!

PR #60, opened 2021-04-07
Allow workflow run timeouts

drewhoskins-stripe:
We were hardcoding the run timeout to be the execution timeout.  This is the default but we should allow people to change it as is possible in the other SDKs.

# Test plan
`bundle exec rspec spec/unit/lib/temporal_spec.rb`
Manual testing:
Launch a docker instance
```
cd examples
./bin/worker&
./bin/trigger TimeoutWorkflow 6 2
```
Screen Shot 2021-04-07 at 11 16 10 AM


robholland (approved):

antstorm (approved):
👍

PR #61, opened 2021-04-09
Continue as new

robholland:


lib/temporal/workflow/context.rb
```diff
@@ -171,12 +176,32 @@ def cancel_timer(timer_id)
...
       end
 
       # TODO: check if workflow can be failed
       def fail(exception)
         command = Command::FailWorkflow.new(exception: exception)
         schedule_command(command)
+        completed!
+      end
+
+      def continue_as_new(workflow_class, *input, **args)
```
antstorm, line 189: I'm assuming most of the times you'd want to "continue as new" the same workflow, right? In which case let's remove `workflow_class` for the arguments and just use `self.class` instead. This can be overridden via passing everything via `options` if needed

antstorm (commented):
Great PR, thank you! One small comment on the interface

lib/temporal/workflow/context.rb
robholland: I'm not able to skip the argument because the context does not appear to know which workflow class it's for. `self.class` is `Temporal::Workflow::Context` not the workflow class whose `execute` we are running.

drewhoskins-stripe (commented):
can you add automated tests?

lib/temporal/workflow/context.rb
drewhoskins-stripe: Should we fix the construction of the context so that it stores the workflow class?

robholland: I prefer it as it is personally as that's the upstream API and matches other SDKs.

antstorm:
@robholland workflow name should be available via `context.metadata.name`. I don't think keeping consistent API between SDKs is something we should worry about — Ruby allows for much more flexibility and thus is already different (ask Maxim for more context on this, he suggested leveraging more dynamism in the first place).

Also specifying the workflow name is redundant and leaves unnecessary room for error

robholland:
I prefer using the class so that configured timeouts and retry policy are applied which will not be the case if we just use the name from the metadata. I will change it though.

antstorm:
@robholland class is better, agree. But it shouldn't be that hard to get workflow class available from within the context (when we initialise the context the workflow class is already available)

robholland:
Updated to store class on the context.

lib/temporal/client/serializer/start_child_workflow.rb
```diff
@@ -1,45 +1,31 @@
...
+              Temporal::Api::Command::V1::StartChildWorkflowExecutionCommandAttributes.new(
                 namespace: object.namespace,
                 workflow_id: object.workflow_id.to_s,
                 workflow_type: Temporal::Api::Common::V1::WorkflowType.new(name: object.workflow_type),
                 task_queue: Temporal::Api::TaskQueue::V1::TaskQueue.new(name: object.task_queue),
                 input: Payload.new(object.input).to_proto,
                 workflow_execution_timeout: object.timeouts[:execution],
                 workflow_task_timeout: object.timeouts[:task],
-                retry_policy: serialize_retry_policy(object.retry_policy),
+                retry_policy: Temporal::Client::Serializer::RetryPolicy.new(object.retry_policy).to_proto,
```
antstorm, line 21: Brilliant 👍

antstorm (approved):

PR #63, opened 2021-04-13
Allow configuring the drain grace period.

robholland:
Defaults to 25 seconds to fit inside the default grace period for
kubernetes pod shutdown which is 30 seconds.

### DIFF
--- a/lib/temporal/worker.rb
+++ b/lib/temporal/worker.rb
@@ -7,9 +7,11 @@ require 'temporal/middleware/entry'
 
 module Temporal
   class Worker
+    DEFAULT_SHUTDOWN_GRACE_PERIOD = 25 # seconds
+
     # activity_thread_pool_size: number of threads that the poller can use to run activities.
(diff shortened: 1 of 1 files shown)


antstorm:
@robholland can you please give me a bit more context on this change? Have you seen issues where 1s isn't enough?

robholland:
It's common for activities to take a longer than 1 second. Particularly during deploys this can result in a lot of activities that have to wait to timeout and be retried. Making this configurable allows people to trade a slower shutdown for less disruption during deploys/restarts.

antstorm:
@robholland we do not stop any workflow/activity tasks that are being processed when killing workers. There's `pollers.each(&:wait)` call that will ensure that we will only exit the process when all the dispatched tasks are fully processed.

This timeout is needed to prevent a rare high load case when a dispatched task gets dropped because the pipe gets closed unsafely (more context [here](https://github.com/temporalio/temporal/issues/1058)).

PR #64, opened 2021-04-20
Allow heartbeats in local testing mode

drewhoskins-stripe:
Unblocks product engineers calling activities that `heartbeat` in local testing mode.

## Test Plan
`bundle exec rspec spec/unit/lib/temporal/testing/local_workflow_context_spec.rb`


antstorm (approved):
👍

PR #66, opened 2021-04-27
Changes to retry policy validation

drewhoskins-stripe:
## Summary
* When there are no retries, it shouldn't be necessary to set the other retry policy parameters.  I think `nil` will be less confusing than a bunch of values that aren't actually used.
* `expiration_interval` is unused
* temporal allows max_attempts to be nil, see https://docs.temporal.io/docs/go/retries/

## Test Plan
New test of a "clean" no-retry policy and an unlimited-attempts policy:
```bundle exec rspec spec/unit/lib/temporal/retry_policy_spec.rb```

### DIFF
--- a/lib/temporal/retry_policy.rb
+++ b/lib/temporal/retry_policy.rb
@@ -1,8 +1,9 @@
 require 'temporal/errors'
 
 module Temporal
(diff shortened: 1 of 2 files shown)


antstorm (approved):
👍

Issue #70, opened 2021-05-07
SSL Support for GRPC Client


tristan-instacart:
Hello.

I have prepared a local branch with SSL support and updated documentation for using it. In my use case we deployed in AWS and needed to run SSL on our load balancer to the frontend service. I was unable to push the branch and open the PR however. 

I'd be happy to open a PR for the change, it's actually fairly simple. 

Additionally, is this SDK going to get revamped to support the Rust based SDK core? Obviously my SSL changes would need some revision if that were the case. 

Thank you for all the work you've put in so far, it's been very helpful to have this SDK out there!

antstorm:
@tristan-instacart glad you're enjoying the SDK!

Feel free to fork the repo and submit a PR via your fork (very common approach in open source, here's a tutorial — https://www.dataschool.io/how-to-contribute-on-github/). I did a test implementation of the SSL at some point, but please feel free to contribute yours as I might not have the time to get to submitting mine this week.

We are not yet planning to switch to using the core SDK, as it's probably still months away. With this in mind an interim SSL solution for the current version would be great

jlemesh:
@tristan-instacart I'll continue if you don't mind.

antstorm:
@tristan-instacart any chance we can finish this one and merge it?

PR #73, opened 2021-05-13
Tristan instacart/add basic ssl support

tristan-instacart:
This Pull Request references issue https://github.com/coinbase/temporal-ruby/issues/70

It is a simple implementation for allowing SSL support by allowing the consumer of the library to pass channel credentials. A very simple use case is added to the README for reference. 

It maintains the default behavior of the SDK for backwards compatibility.


tristan-instacart:
Investigating test failure

tristan-instacart:
Sorry for the test error, copy paste issue from a monkeypatch version I based the PR on, fixed now.

lib/temporal/client/grpc_client.rb
```diff
@@ -379,7 +379,7 @@ def cancel_polling_request
       def client
         @client ||= Temporal::Api::WorkflowService::V1::WorkflowService::Stub.new(
           url,
-          :this_channel_is_insecure,
+          Temporal.configuration.grpc_ssl_config,
```
antstorm, line 382: Would you mind passing this as the argument (similar to host/port) to the client to avoid this extra dependency?

lib/temporal/configuration.rb
```diff
@@ -26,12 +26,13 @@ class Configuration
 
     def initialize
       @client_type = :grpc
-      @logger = Temporal::Logger.new(STDOUT, progname: 'temporal_client')
+      @logger = Logger.new(STDOUT, progname: 'temporal_client')
```
antstorm, line 29: What is the reason for this change?

tristan-instacart: This is a mistake, it is legacy from the code change existing before Temporal::Logger. 
I will remove it.

lib/temporal/client/grpc_client.rb
tristan-instacart: Still pull from config but do so here yeah?
 https://github.com/coinbase/temporal-ruby/blob/a7665ee115610f5eb6004d3dc92fbd297ef0e300/lib/temporal/client.rb#L18

tristan-instacart:
Refactored, then tested against an env in AWS behind an SSL enabled ALB to ensure it is connecting via SSL successfully.

tristan-instacart:
And now we have learned the valuable lesson of running the rspec locally :)

tristan-instacart:
Refactored, then tested against an env in AWS behind an SSL enabled ALB to ensure it is connecting via SSL successfully.

antstorm:
@tristan-instacart can you please rebase it against the master? There are a few conflict, but otherwise looking good to go 👍

lib/temporal/client.rb
```diff
@@ -15,7 +15,11 @@ def self.generate
       thread_id = Thread.current.object_id
       identity = "#{thread_id}@#{hostname}"
 
-      client_class.new(host, port, identity)
+      if Temporal.configuration.client_type == :grpc
+        client_class.new(host, port, identity, Temporal.configuration.grpc_ssl_config)
+      else
+        client_class.new(host, port, identity)
+      end
```
antstorm, line 22: Can you please add a test case for this?

lib/temporal/client/grpc_client.rb
```diff
@@ -388,12 +389,12 @@ def cancel_polling_request
...
       private
 
-      attr_reader :url, :identity, :poll_mutex, :poll_request
+      attr_reader :url, :channel_creds, :identity, :poll_mutex, :poll_request
 
       def client
         @client ||= Temporal::Api::WorkflowService::V1::WorkflowService::Stub.new(
           url,
-          :this_channel_is_insecure,
+          channel_creds,
```
antstorm, line 397: And this?

README.md
```diff
@@ -55,6 +55,7 @@ Temporal.configure do |config|
   config.port = 7233
   config.namespace = 'ruby-samples'
   config.task_queue = 'hello-world'
+  config.channel_creds = :this_channel_is_insecure
```
antstorm, line 58: nit: Ruby folks are ok with full names, let's probably call this just `credentials` unless you think it might conflict with some other config we'd want to add in the future

lib/temporal/configuration.rb
```diff
@@ -50,5 +51,13 @@ def task_list=(name)
     def timeouts=(new_timeouts)
       @timeouts = DEFAULT_TIMEOUTS.merge(new_timeouts)
     end
+
+    def grpc_ssl_config
+      if @channel_creds.nil?
+        :this_channel_is_insecure
```
antstorm, line 57: also would be great to have a spec for this and also put the value in a constant

PR #76, opened 2021-05-21
[WIP] Correctly handle timer/activity cancelation events

### LINKED ISSUE #75: Unexpected non-determinism error when using timers
I'm running into an unexpected non-determinism error. 

### Repro steps
1. Modify the `CancellingTimerWorkflow` to be:
```
class CancellingTimerWorkflow < Temporal::Workflow
  def execute(activity_timeout, timer_timeout)
    activity_future = SleepActivity.execute(activity_timeout)
    timer = workflow.start_timer(timer_timeout)
    timer.done do
      if !activity_future.finished?
        HelloWorldActivity.execute!('reminder')
      end
    end

    activity_future.wait
    timer.cancel
    HelloWorldActivity.execute!('done')
  end
end
```
This change is on the latest commit of https://github.com/jeffschoner-stripe/temporal-ruby/tree/non-determinism-repro if you'd prefer to check out the branch.

2. Execute the workflow with an `activity_timeout` less than the `timer_timeout` such that the timer will be canceled. E.g.,
```
examples/bin/trigger CancellingTimerWorkflow 5 15
```

### Expected Behavior
`SleepActivity` completes after 5 seconds, timer gets canceled, workflow completes successfully

### Observed behavior
`SleepActivity` completes after 5 seconds, but the workflow task fails with a message of "Unexpected command cancel_timer_request"

Captured history (JSON and screenshot) of my execution of this workflow:
[workflow history JSON export](https://github.com/coinbase/temporal-ruby/files/6511994/dfad7d5b.a1c5.4a62.b935.6d8645714d04.-.eac20293-b4cd-41a7-a81f-b9a507769f14.txt)

Screen Shot 2021-05-19 at 16 33 38

### Other Notes
- Originally, I was seeing this when the `SleepActivity` was replaced by an asynchronously completed activity, but that does not seem to matter as it reproduces with both kinds of activities. The sleep is easier to execute than triggering the async complete.
- If the `activity_timeout` is larger than the `timer_timeout`, the workflow executes as expected. The timer fires before the `SleepActivity` has completed, its block executes, and the timer is canceled once the activity does complete.

antstorm:
Fixes #75 

The non-determinism check depends on the class `EventTarget` that creates a canonical event that is used when matching commands to history events (schedule_activity -> activity_scheduled, etc). This allows to check the pending command (not yet sent to Temporal) and history events and remove the ones that are already in the history (in the same order as they were scheduled).

There were 2 bugs with this:
1. When creating an `EventTarget` from an event we did a prefix match favouring the first match. This means that between `ACTIVITY_TASK` and `ACTIVITY_TASK_CANCEL` the first one would win and will generate an incorrect event target (activity, instead of cancel_activity). Fixed by matching all and picking the last match.
2. TIMER_CANCELED event (along with ACTIVITY_TASK_CANCELED) was not discarding a pending command, which it should have. This should be fixed along with `REQUEST_CANCEL_ACTIVITY_TASK_FAILED` and `CANCEL_TIMER_FAILED` which should also remove the cancelation command from the pending list

How was it tested?
Manually by running workflows, proper specs pending

### DIFF
--- a/lib/temporal/workflow/history/event_target.rb
+++ b/lib/temporal/workflow/history/event_target.rb
@@ -17,10 +17,12 @@ module Temporal
         WORKFLOW_TYPE                         = :workflow
         CANCEL_WORKFLOW_REQUEST_TYPE          = :cancel_workflow_request
 
+        # NOTE: The order is important, first prefix match wins (should correspond to a longer match)
         TARGET_TYPES = {
-          'ACTIVITY_TASK'                              => ACTIVITY_TYPE,
           'ACTIVITY_TASK_CANCEL'                       => CANCEL_ACTIVITY_REQUEST_TYPE,
+          'ACTIVITY_TASK'                              => ACTIVITY_TYPE,
           'REQUEST_CANCEL_ACTIVITY_TASK'               => CANCEL_ACTIVITY_REQUEST_TYPE,
+          'TIMER_CANCELED'                             => CANCEL_TIMER_REQUEST_TYPE,
           'TIMER'                                      => TIMER_TYPE,
           'CANCEL_TIMER'                               => CANCEL_TIMER_REQUEST_TYPE,
           'CHILD_WORKFLOW_EXECUTION'                   => CHILD_WORKFLOW_TYPE,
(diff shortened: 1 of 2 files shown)


lib/temporal/workflow/history/event_target.rb
```diff
@@ -42,7 +44,7 @@ def self.workflow
         end
 
         def self.from_event(event)
-          _, target_type = TARGET_TYPES.find { |type, _| event.type.start_with?(type) }
+          _, target_type = TARGET_TYPES.find_all { |type, _| event.type.start_with?(type) }.last
```
robholland, line 47: Why not sort the keys by length at load time to avoid having match multiple times?

antstorm: Yep, that does make more sense

antstorm:
Fixed here — https://github.com/coinbase/temporal-ruby/pull/96/files

PR #74, opened 2021-05-14
Retryer; Retries on activity completion/failure

drewhoskins-stripe:
# Description
* Java SDK's workers retry when trying to report activity completion/failure.  This should increase reliability when the server is down.
* I tried to use the exact backoff policy parameters they use for their default [GrpcRetryer](https://github.com/temporalio/sdk-java/blob/ad8831d4a4d9d257baf3482ab49f1aa681895c0e/temporal-sdk/src/main/java/io/temporal/internal/common/GrpcRetryer.java#L36), see [RpcRetryOptions](https://github.com/temporalio/sdk-java/blob/ad8831d4a4d9d257baf3482ab49f1aa681895c0e/temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java#L32)

It will retry for a minute.  Note that if the retries are longer than the activity timeout, it will be hitting temporal-server fruitlessly.  Temporal handles this gracefully by complaining that the workflow has already finished.  We could in theory try to incorporate the activity timeout into the limit calculation, but I don't think the Java SDK does.

* SleepActivity was useless with such short timeouts.

# Test Plan
Unit test of the retryer logic:
`rspec spec/unit/lib/temporal/client/retryer.rb`
Manual:
```
cd examples
./bin/worker
# other terminal
./bin/trigger CancellingTimerWorkflow 15 1
```
Kill temporal-server before 15 seconds has elapsed, then restart the server.  Notice the retries followed by success:
```
I, [2021-05-14T09:29:10.263760 #68971]  INFO -- temporal_client: [EXAMPLE]: Finished {"metadata":{"namespace":"ruby-samples","workflow_id":"cbf5955b-c1e2-4806-a8da-248052711793","workflow_name":"CancellingTimerWorkflow","workflow_run_id":"e688fe0c-cf7c-4bd3-8fe4-81b42bc17638","activity_id":"6","activity_name":"SleepActivity","attempt":1}}
I, [2021-05-14T09:29:10.263905 #68971]  INFO -- temporal_client: Activity task completed {"namespace":"ruby-samples","workflow_id":"cbf5955b-c1e2-4806-a8da-248052711793","workflow_name":"CancellingTimerWorkflow","workflow_run_id":"e688fe0c-cf7c-4bd3-8fe4-81b42bc17638","activity_id":"6","activity_name":"SleepActivity","attempt":1}
[...]

PR #77, opened 2021-06-02
Speed up unit tests that hit the Retryer

drewhoskins-stripe:
## Description
https://github.com/coinbase/temporal-ruby/pull/74 regressed the speed because we added retries and `sleep` to certain code paths.

## Test Plan
Observe that this executes quickly:
```
GitHub/temporal-ruby $ bundle exec rspec spec/unit/
...................................................................................................................................................................................................................................................................................

Finished in 1.08 seconds (files took 0.9169 seconds to load)
275 examples, 0 failures
```

### DIFF
--- a/spec/unit/lib/temporal/activity/task_processor_spec.rb
+++ b/spec/unit/lib/temporal/activity/task_processor_spec.rb
@@ -35,6 +35,9 @@ describe Temporal::Activity::TaskProcessor do
       allow(middleware_chain).to receive(:invoke).and_call_original
 
       allow(Temporal.metrics).to receive(:timing)
+
+      # Skip sleeps during retries to speed up the test.
+      allow(Temporal::Client::Retryer).to receive(:sleep).and_return(nil)
     end
 
     context 'when activity is not registered' do


spec/unit/lib/temporal/activity/task_processor_spec.rb
```diff
@@ -53,6 +53,8 @@
         allow(client)
           .to receive(:respond_activity_task_failed)
           .and_raise(StandardError)
+        # Skip sleeps during retries to speed up the test.
+        expect(Temporal::Client::Retryer).to receive(:sleep).at_least(1).times
```
antstorm, line 57: nit: we're trying not to use `expect().to receive()` construct in specs because it conflates two thing — stubbing and checking. Instead all the spec explicitly stub via `allow().to receive()` and then check with `expect().to have_received()`

spec/unit/lib/temporal/activity/task_processor_spec.rb
```diff
@@ -107,6 +109,8 @@
           allow(client)
             .to receive(:respond_activity_task_completed)
             .and_raise(StandardError)
+          # Skip sleeps during retries to speed up the test.
+          expect(Temporal::Client::Retryer).to receive(:sleep).at_least(1).times
```
antstorm, line 113: Also, I think for this spec you don't care if `:sleep` was called, but you do care that `:with_retries` was. Otherwise you're lurking into `Retryer`'s implementation details and this would break if you change `sleep` for something else, e.g. `Kernel.sleep`

antstorm (commented):
This makes sense 👍 

Btw, should we default to no-op `sleep` in `Temporal::Client::Retryer` for all specs? Something like `allow(Temporal::Client::Retryer).to receive(:sleep).and_return(nil)`?

This way all the future tests that might be hitting this path do not need to stub it

antstorm (approved):

PR #78, opened 2021-06-04
Polling for workflow completion and results

drewhoskins-stripe:
## Description
This allows you to synchronously long poll for a workflow to complete and then either
(a) return the result returned from execute(), or
(b) raise an exception if the workflow failed

Also, found a bug in `terminate_workflow` caused by the recent serialization change.

## Test Plan
Integration tests
Terminal 1
```
cd examples
./bin/worker
```
Terminal 2
```
USE_ENCRYPTION=1 ./bin/worker
```
Terminal 3
```
cd examples
bundle exec rspec -f d spec/integration/await_workflow_result.rb
bundle exec rspec -f d spec/integration/terminate_workflow.rb
```

Unit: 
```
rspec ./spec/unit/lib/temporal_spec.rb
```


lib/temporal.rb
```diff
@@ -81,6 +83,40 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
       )
     end
 
+    def await_workflow_result(workflow:, workflow_id:, run_id:, **args)
```
drewhoskins-stripe, line 86: some of the other "follow up" methods take a `namespace` instead of a workflow, but I found it hard to use.  I wanted it to match the same logic as `start_workflow`.  It's actually harder for developers to get the namespace and replicate the logic themselves.  Would love to see if we can go back and fix the others.
This is also partially inconsistent in that it takes kwargs.  I noted that some of other methods take the same parameters in different orders and with different choices on keyword vs positional args.  just going full-on kwargs seems to be the best policy, and easier to read the callsites.

lib/temporal/errors.rb
```diff
@@ -25,6 +25,24 @@ class WorkflowNotRegistered < ClientError; end
   class ApiError < Error; end
 
   class NotFoundFailure < ApiError; end
+
+  # Superclass for errors raised when retrieving a workflow result on the
```
jeffschoner-stripe, line 29: nit: These are technically broader than what's described here. A terminated workflow, for example, doesn't involve any failure on the worker.

examples/workflows/randomly_failing_workflow.rb
```diff
@@ -0,0 +1,11 @@
+require 'activities/randomly_failing_activity'
+
+class RandomlyFailingWorkflow < Temporal::Workflow
+  def execute
+    RandomlyFailingActivity.execute!
+
+    return 'You are very lucky!'
+  rescue RandomlyFailingActivity::WrongGuess => e
+    return e.message
```
jeffschoner-stripe, line 9: You can omit `return`

lib/temporal.rb
jeffschoner-stripe, line 86: Can `run_id` be omitted or specified as `nil` to indicate the latest run of the workflow should be used? If so, it's worth calling out that behavior in a comment.

drewhoskins-stripe: I'll add a test and find out!

examples/workflows/randomly_failing_workflow.rb
drewhoskins-stripe: This is a copy/paste job.  The house style seems to be to not emit.

lib/temporal.rb
```diff
@@ -81,6 +83,41 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
       )
     end
 
+    # run_id of nil: await the latest run
+    def await_workflow_result(workflow:, workflow_id:, run_id: nil, **args)
```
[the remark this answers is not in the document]
drewhoskins-stripe: @antstorm I don't consider this highly specialized.  It'll be used in integration tests frequently, and sometimes in production.  In particular, this scenario will be more common than checking history.
A couple other points.
- The fact that this happens to use the underlying history GRPC endpoint is an implementation detail.  It could just as easily have been in "DescribeWorkflow" or a dedicated endpoint.  People would struggle to figure out that that's the right thing to call when wanting to check the result.
- The complexity of this method tells me that we wouldn't want developers to need to reimplement the logic.

spec/unit/lib/temporal/await_workflow_result.rb
```diff
@@ -0,0 +1,50 @@
...
+  class NamespacedWorkflow < Temporal::Workflow
+    namespace 'some-namespace'
+    task_queue 'some-task-queue'
+  end
+
+  let(:client) { instance_double(Temporal::Client::GRPCClient) }
+  before { allow(Temporal::Client).to receive(:generate).and_return(client) }
+  after { Temporal.remove_instance_variable(:@client) rescue NameError }
+
+  let(:response) do
```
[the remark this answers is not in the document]
drewhoskins-stripe: There are integration tests.

examples/spec/integration/await_workflow_result.rb
```diff
@@ -0,0 +1,124 @@
...
+    # old run ID still works
+    actual_old_result = Temporal.await_workflow_result(
+      workflow: ResultWorkflow,
+      workflow_id: workflow_id,
+      run_id: first_run_id,
+    )
+    expect(actual_old_result).to eq(expected_first_result)
+  end
+
+  it 'is idempotent' do
```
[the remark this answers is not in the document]
drewhoskins-stripe: Yeah, I wanted to make sure that it was idempotent with the combination of parameters I used.

lib/temporal/client/grpc_client.rb
```diff
@@ -311,7 +311,7 @@ def terminate_workflow_execution(
             run_id: run_id,
           ),
           reason: reason,
-          details: to_details_payload(details)
+          details: to_details_payloads(details)
```
drewhoskins-stripe, line 314: Trivial bug fix for the recent serialization change.  It's tested in the new terminate integration test.

examples/spec/integration/serial_hello_world_workflow_spec.rb
```diff
@@ -4,13 +4,11 @@
   it 'completes' do
     workflow_id, run_id = run_workflow(described_class, 'Alice', 'Bob', 'John')
 
-    result = fetch_history(
-      workflow_id,
-      run_id,
-      wait_for_new_event: true,
-      event_type: :close,
+    result = Temporal.await_workflow_result(
```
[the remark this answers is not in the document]
drewhoskins-stripe: 👍 , I suppose for *our* specs it's not that useful because we're not really testing product behavior.
It is a good example for integrating teams to copy though.

antstorm (commented):
One question I have is around raising errors in response to this method call. When do you think it might be useful? Another approach might be to return a result object that holds a result or an error for failures

spec/unit/lib/temporal_spec.rb
```diff
@@ -362,6 +362,99 @@ class TestStartWorkflow < Temporal::Workflow
...
+      end
+
+      [
+        {type: 'hash', expected_result: { 'key' => 'value' }},
+        {type: 'integer', expected_result: 5},
+        {type: 'nil', expected_result: nil},
+        {type: 'string', expected_result: 'a result'},
+      ].each do |type:, expected_result:|
+        it "completes and returns a #{type}" do
+          payload = Temporal::Api::Common::V1::Payloads.new(
```
drewhoskins-stripe, line 403: I tried to add a Fabricator but got weird errors and had to timebox the exercise. :/

jeffschoner-stripe (approved):

lib/temporal.rb
```diff
@@ -81,6 +83,58 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
...
+          )
+        when :EVENT_TYPE_WORKFLOW_EXECUTION_CONTINUED_AS_NEW
+          new_run_id = event['workflow_execution_continued_as_new_event_attributes']['new_execution_run_id']
+          if run_id
+            # If they specified a run ID, we should throw to let them know they're not getting the result
+            # they wanted.  They can re-call on the new run ID if they want.
+            raise Temporal::WorkflowRunContinuedAsNew.new(new_run_id: new_run_id)
+          else
+            current_run_id = new_run_id
+            # await the next run until the workflow is complete.
```
antstorm, line 130: As far as I understand a very common use-case for ContinueAsNew is periodic execution that might actually never stop. We might be sending people into an endless loop here

lib/temporal.rb
```diff
@@ -81,6 +83,58 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
...
+      current_run_id = run_id
+      loop do
+        history_response = client.get_workflow_execution_history(
+          namespace: execution_options.namespace,
+          workflow_id: workflow_id,
+          run_id: current_run_id,
+          wait_for_new_event: true,
+          event_type: :close
+        )
+        event = history_response['history']['events'].first
```
antstorm, line 103: Let's use the existing Workflow::History class here instead of an ad-how history parsing

lib/temporal.rb
```diff
@@ -81,6 +83,58 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
...
+        case event.event_type
+        when :EVENT_TYPE_WORKFLOW_EXECUTION_COMPLETED
+          payloads = event['workflow_execution_completed_event_attributes'].result
+          return nil if !payloads # happens when the workflow itself returns nil
+          return Temporal::Client::Converter::Payload::JSON.new.from_payload(payloads['payloads'].first)
```
antstorm, line 108: This is peeking into too low level API and also ignores any pre-set converters (such as encryption, etc). We need to use something higher level. @robholland can you please advice?

lib/temporal.rb
```diff
@@ -81,6 +83,58 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
...
+          raise Temporal::WorkflowTimedOut
+        when :EVENT_TYPE_WORKFLOW_EXECUTION_TERMINATED
+          raise Temporal::WorkflowTerminated
+        when :EVENT_TYPE_WORKFLOW_EXECUTION_CANCELED
+          raise Temporal::WorkflowCanceled
+        when :EVENT_TYPE_WORKFLOW_EXECUTION_FAILED
+          # failure_info: Temporal::Api::Failure::V1::Failure
+          failure_info = event['workflow_execution_failed_event_attributes']['failure']
+          raise Temporal::WorkflowFailed.new(
+            failure_info['message'],
```
antstorm, line 119: We most likely need the actual error class of the failure here for continuity. Message will not contain that

lib/temporal.rb
drewhoskins-stripe: We may, it's true, though I would hope that it'd be quite rare for a developer to make that mistake.
In our wrappers at Stripe, we can easily detect that it's a cron-scheduled workflow and barf if somebody awaits it without a run_id.
In stock temporal-ruby, I thought about adding a `timeout` parameter, but that'd involve spinning up a thread and I wasn't sure if it would be better to just let the client do that part.  Still, I'd be happy to add that in a followup PR and then I think that largely solves this.

lib/temporal.rb
robholland: https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/workflow/history.rb#L9

robholland:
The child workflow code I was referring to is: https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/workflow/history.rb#L9 I'd need to dig in to see what that actually raises in the various circumstances.

lib/temporal.rb
robholland: The payload concern (https://github.com/coinbase/temporal-ruby/blob/75a3f854557ed47cbbdd4d115e40903b28b55d3b/lib/temporal/concerns/payloads.rb) should be used here: `from_result_payloads(payloads)`.

lib/temporal.rb
```diff
@@ -81,6 +83,59 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
       )
     end
 
+    # Long polls for a workflow to be completed and returns whatever the execute function
+    # returned.
+    # run_id of nil: await the entire workflow completion.  This can span multiple runs
+    # in the case where the workflow uses continue as new.
+    def await_workflow_result(workflow, workflow_id:, run_id: nil, **args)
```
drewhoskins-stripe, line 90: As far as refining the interface, it seems important that we use the word `await` (ish) so that people know this is a blocking call.
* I looked futures for activities, but it's all built into WorkflowContext and can't be used here.
* The assumption inside workflows is that things are blocking.  The assumption outside is that they are not.  So I don't think we should share the naming or approach we use when awaiting activities.

lib/temporal.rb
```diff
@@ -81,6 +83,59 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
...
+        history = Workflow::History.new(history_response.history.events)
+        closed_event = history.events.first
+        case closed_event.type
+        when 'WORKFLOW_EXECUTION_COMPLETED'
+          payloads = closed_event.attributes.result
+          return nil if !payloads # happens when the workflow itself returns nil
+          return from_result_payloads(payloads)
+        when 'WORKFLOW_EXECUTION_TIMED_OUT'
+          raise Temporal::WorkflowTimedOut
+        when 'WORKFLOW_EXECUTION_TERMINATED'
```
drewhoskins-stripe, line 112: https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/workflow/state_manager.rb#L320-L339 is the closest thing to share with, it has some advantages and limitations.
The most important bit for consistency is to agree on whether we prefer Temporal::WorkflowTimedOut or just Temporal::TimedOut .
I prefer the former for consistency and clarity (and I like that it shares a base class with the others), but not sure if I'm allowed to make that change.  @antstorm thoughts?

I'd like to share code but it'll be non-trivial to test the existing state manager using the new code, so treating that as a lower priority.

lib/temporal.rb
drewhoskins-stripe: We can easily add a non-blocking version of this that returns a future later on.

lib/temporal/workflow/errors.rb
```diff
@@ -0,0 +1,39 @@
+module Temporal
+  class Workflow
+    class Errors
+      include Concerns::Payloads
+
+      # Convert a failure returned from the server to an Error to raise to the client
+      # failure: Temporal::Api::Failure::V1::Failure
+      def error_from(failure, default_exception_class = StandardError)
+        case failure.failure_info
```
drewhoskins-stripe, line 9: Copied verbatim from the state manager.  This could be a good candidate for unit tests if you want to beef up coverage and correctness here.

lib/temporal/workflow/errors.rb
```diff
@@ -0,0 +1,39 @@
+module Temporal
+  class Workflow
+    class Errors
+      include Concerns::Payloads
+
+      # Convert a failure returned from the server to an Error to raise to the client
+      # failure: Temporal::Api::Failure::V1::Failure
+      def error_from(failure, default_exception_class = StandardError)
```
drewhoskins-stripe, line 8: naming similar to many other `_from` methods.

lib/temporal.rb
```diff
@@ -81,6 +84,54 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
...
+          run_id: current_run_id,
+          wait_for_new_event: true,
+          event_type: :close
+        )
+        history = Workflow::History.new(history_response.history.events)
+        closed_event = history.events.first
+        case closed_event.type
+        when 'WORKFLOW_EXECUTION_COMPLETED'
+          payloads = closed_event.attributes.result
+          return nil if !payloads # happens when the workflow itself returns nil
```
antstorm, line 109: `from_result_payloads` should be able to handle this use case just fine, no need to special case it

lib/temporal.rb
```diff
@@ -81,6 +84,54 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
...
+          return from_result_payloads(payloads)
+        when 'WORKFLOW_EXECUTION_TIMED_OUT'
+          raise Temporal::WorkflowTimedOut
+        when 'WORKFLOW_EXECUTION_TERMINATED'
+          raise Temporal::WorkflowTerminated
+        when 'WORKFLOW_EXECUTION_CANCELED'
+          raise Temporal::WorkflowCanceled
+        when 'WORKFLOW_EXECUTION_FAILED'
+          raise Temporal::Workflow::Errors.new.error_from(closed_event.attributes.failure)
```
antstorm, line 118: no need to initialize an object there, can just be a class method instead `Errors.generate_from()`

lib/temporal.rb
antstorm: I would much rather not handle that case at all and remove the loop until we find a better way

lib/temporal.rb
```diff
@@ -160,6 +211,8 @@ def metrics
 
     private
 
+    include Concerns::Payloads
```
antstorm, line 214: In Ruby we normally place includes at the top of the class definition. The scope of the included method is defined by the module that is being included, not where you include it

lib/temporal/workflow/errors.rb
```diff
@@ -0,0 +1,39 @@
...
+          exception_class = safe_constantize(failure.application_failure_info.type)
+          exception_class ||= default_exception_class
+          message = from_details_payloads(failure.application_failure_info.details)
+          backtrace = failure.stack_trace.split("\n")
+
+          exception_class.new(message).tap do |exception|
+            exception.set_backtrace(backtrace) if !backtrace.empty?
+          end
+        when :timeout_failure_info
+          TimeoutError.new("Timeout type: #{failure.timeout_failure_info.timeout_type.to_s}")
```
antstorm, line 20: We probably want `Temporal::TimeoutError` here, not Ruby's generic one

spec/unit/lib/temporal_spec.rb
```diff
@@ -362,6 +362,99 @@ class TestStartWorkflow < Temporal::Workflow
...
+        )
+      end
+
+      [
+        {type: 'hash', expected_result: { 'key' => 'value' }},
+        {type: 'integer', expected_result: 5},
+        {type: 'nil', expected_result: nil},
+        {type: 'string', expected_result: 'a result'},
+      ].each do |type:, expected_result:|
+        it "completes and returns a #{type}" do
```
antstorm, line 402: Does this test make sense here? It's not testing `await_workflow_result`, but rather how converters work

spec/unit/lib/temporal_spec.rb
```diff
@@ -362,6 +362,99 @@ class TestStartWorkflow < Temporal::Workflow
...
+          payload = Temporal::Api::Common::V1::Payloads.new(
+            payloads: [
+              Temporal::Client::Converter::Payload::JSON.new.to_payload(expected_result)
```
antstorm, line 405: You're forcing JSON converter here, however things such as `nil` and `string` are using other converters

lib/temporal.rb
```diff
@@ -81,6 +84,54 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil)
       )
     end
 
+    # Long polls for a workflow to be completed and returns whatever the execute function
+    # returned.
+    # run_id of nil: await the entire workflow completion.  This can span multiple runs
+    # in the case where the workflow uses continue as new.
```
antstorm, line 90: Let's add a note that this is an unstable API and we might change it to give us some flexibility

lib/temporal/workflow/errors.rb
```diff
@@ -0,0 +1,39 @@
+module Temporal
+  class Workflow
+    class Errors
```
antstorm, line 3: This class needs unit tests

antstorm (changes requested):

spec/unit/lib/temporal_spec.rb
drewhoskins-stripe: It also tests how the converters are called.

lib/temporal/workflow/errors.rb
antstorm: It only does so if you expect the errors to be `require`d in the global context, which is a bad practise. This file does not `require` them, so therefore not guaranteed to resolve it

lib/temporal.rb
drewhoskins-stripe: As I integrated the timeouts, I realized you were right.  The loop is too complicated for what we get out of it.

lib/temporal.rb
```diff
@@ -158,6 +216,11 @@ def metrics
       @metrics ||= Metrics.new(configuration.metrics_adapter)
     end
 
+    class ResultConverter
```
drewhoskins-stripe, line 219: For those curious, I used a combination of responses from here: https://stackoverflow.com/questions/39716799/how-to-implement-private-inner-class-in-ruby

lib/temporal/client/grpc_client.rb
```diff
@@ -124,8 +124,20 @@ def get_workflow_execution_history(
...
         wait_for_new_event: false,
-        event_type: :all
+        event_type: :all,
+        timeout_s: nil
       )
+        if wait_for_new_event 
+          if timeout_s.nil?
+            # This is an internal error.  Wrappers should enforce this.
+            raise "You must specify a timeout when wait_for_new_event = true."
+          elsif timeout_s > 30
```
antstorm, line 134: Can you please move 30 into a constant?

antstorm (changes requested):
Looks great, last few bits!

antstorm (approved):
Thank you!

PR #80, opened 2021-06-10
Fix namespace creation issues in CI.

robholland:
Temporal does not currently offer a way to ensure that a namespace registration is available on all shards.

https://community.temporal.io/t/how-to-tell-the-namespace-is-ready-to-be-polled-by-workers/1606

Unconditionally wait 15 seconds (which is the shard cache refresh time) after namespace registration to ensure it's ready when the workers attempt to use it.


antstorm (approved):
Brilliant, thank you Rob!

PR #81, opened 2021-06-16
Fix serialization of retry policy

drewhoskins-stripe:
Fixes a bug in the recent serialization changes when non-retriable errors are specified.

## Test Plan
`rspec spec/unit/lib/temporal/client/serializer/retry_policy_spec.rb`


robholland (approved):
This was a leftover from the cadence fork.

Fix looks good, thanks!

antstorm (approved):
Thank you!

PR #83, opened 2021-06-22
Special case signal payloads.

robholland:
Other SDKs only use the first payload, as for details/results.


robholland:
h/t weissburg-stripe for reporting, thanks!

antstorm (approved):
Looks good, thank you for fixing this!

PR #84, opened 2021-06-22
Fix deserialization of activity exceptions

drewhoskins-stripe:
We were serializing activity exception messages as arrays of a string rather than a string.

# Test Plan
```
GitHub/temporal-ruby/examples $ bundle exec rspec spec/integration/call_failing_activity_workflow_spec.rb
```


lib/temporal/workflow/state_manager.rb
```diff
@@ -322,10 +322,9 @@ def parse_failure(failure, default_exception_class = StandardError)
         when :application_failure_info
           exception_class = safe_constantize(failure.application_failure_info.type)
           exception_class ||= default_exception_class
-          details = from_payloads(failure.application_failure_info.details)
+          message = from_details_payloads(failure.application_failure_info.details)
           backtrace = failure.stack_trace.split("\n")
-
```
antstorm, line 327: nit: You don't like newlines? :)

lib/temporal/workflow/state_manager.rb
```diff
@@ -322,10 +322,9 @@ def parse_failure(failure, default_exception_class = StandardError)
         when :application_failure_info
           exception_class = safe_constantize(failure.application_failure_info.type)
           exception_class ||= default_exception_class
-          details = from_payloads(failure.application_failure_info.details)
+          message = from_details_payloads(failure.application_failure_info.details)
```
antstorm, line 325: nit: the details might contain more than just a message, it can be a serialized error attributes depending on the error class

examples/spec/integration/call_failing_activity_workflow_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'workflows/call_failing_activity_workflow'
+
+describe CallFailingActivityWorkflow, :integration do
+
+  class TestDeserializer
+    include Temporal::Concerns::Payloads
+  end
+
+  it 'correctly re-raises an activity-thrown exception in the workflow' do
```
antstorm, line 9: this is not a very clean spec because it does not isolate activity -> workflow error serialisation. Can we add a unit test for this specific part?

examples/workflows/call_failing_activity_workflow.rb
```diff
@@ -0,0 +1,14 @@
+require 'activities/failing_activity'
+
+# exercises a workflow's ability to handle errors in user code, which get
+# re-raised in the workflow.
+class CallFailingActivityWorkflow < Temporal::Workflow
+  def execute(message)
+    begin
+      FailingActivity.execute!(message)
+    rescue => e
+      return {class: e.class, message: e.message}
```
antstorm, line 10: Why not just re-raise the error? I'm assuming it will expose the same issue and is a bit clearer

antstorm:
@robholland thoughts?

examples/spec/integration/call_failing_activity_workflow_spec.rb
drewhoskins-stripe: (a) There are no existing unit tests of lib/temporal/workflow/state_manager.rb so it's not easy to test in isolation, and 
(b) I need to make sure that the serialization matches the deserialization.  I want a test that encompasses both phases.

lib/temporal/workflow/state_manager.rb
drewhoskins-stripe: @antstorm I thought/hoped the same at first, but alas, we only serialize the message.
https://github.com/coinbase/temporal-ruby/blob/7d3f1bf82b434af54d2f39956a4263433c9719ed/lib/temporal/client/serializer/failure.rb#L16

lib/temporal/workflow/state_manager.rb
drewhoskins-stripe: inadvertent

drewhoskins-stripe:
@antstorm , Rob already looked at this in a previous revision.

examples/workflows/call_failing_activity_workflow.rb
drewhoskins-stripe: Agreed.  Once my other PR is in, it'll be easier to test that, as it's a bit difficult to extract the exception from the payload.  I can hold off til that's in.

lib/temporal/workflow/state_manager.rb
antstorm: Yeah, something we'll need to fix

examples/spec/integration/call_failing_activity_workflow_spec.rb
antstorm: (a) I'll bootstrap specs for the state manager, so you can piggy back
(b) The main change here is exposing the original error to a workflow. The spec you wrote adds a complexity of workflow result serialization/deserialization using a very different format

drewhoskins-stripe: @antstorm There's a circular dependency.  If you accept this, the test and sample are fixed in https://github.com/coinbase/temporal-ruby/pull/78 .
That's rebased on top of this PR which fixes this test and makes it very clean.
It also shares the same deserialization code at the Activity call as well as polling for the workflow result.
Both changes:
https://github.com/coinbase/temporal-ruby/pull/78/files/b22190eec6d1e2b34d4c2a762a4f946f72ed3b6b..ccdf2457dfee684e237902653a50ff6f9d1e724c

examples/workflows/call_failing_activity_workflow.rb
drewhoskins-stripe: See comment above.

antstorm (approved):

PR #86, opened 2021-06-29
[Refactor] Improve Retryer specs

antstorm:
A few improvements:

- Turn Retryer.do_not_retry_errors into a frozen constant
- Make Retryer specs a bit clearer

No functional changes

### DIFF
--- a/lib/temporal/client/retryer.rb
+++ b/lib/temporal/client/retryer.rb
@@ -9,19 +9,17 @@ module Temporal
       # List pulled from RpcRetryOptions in the Java SDK
       # https://github.com/temporalio/sdk-java/blob/ad8831d4a4d9d257baf3482ab49f1aa681895c0e/temporal-serviceclient/src/main/java/io/temporal/serviceclient/RpcRetryOptions.java#L32
       # No amount of retrying will help in these cases.
-      def self.do_not_retry_errors
-        [
-          GRPC::AlreadyExists,
-          GRPC::Cancelled,
-          GRPC::FailedPrecondition,
-          GRPC::InvalidArgument,
-          # If the activity has timed out, the server will return this and will never accept a retry
-          GRPC::NotFound,
-          GRPC::PermissionDenied,
-          GRPC::Unauthenticated,
-          GRPC::Unimplemented,
-        ]
-      end
+      DO_NOT_RETRY_ERRORS = [
+        GRPC::AlreadyExists,
+        GRPC::Cancelled,
+        GRPC::FailedPrecondition,
+        GRPC::InvalidArgument,
+        # If the activity has timed out, the server will return this and will never accept a retry
+        GRPC::NotFound,
+        GRPC::PermissionDenied,
+        GRPC::Unauthenticated,
+        GRPC::Unimplemented,
+      ].freeze
 
       # Used for backoff retries in certain cases when calling temporal server.
       # on_retry - a proc that's executed each time you need to retry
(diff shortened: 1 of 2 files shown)


spec/unit/lib/temporal/client/retryer_spec.rb
```diff
@@ -78,37 +78,30 @@
...
-      on_retry = Proc.new {
-        on_retry_calls += 1
-      }
-
-      begin
-        described_class.with_retries(on_retry: on_retry) do
-          if !retried
-            retried = true
-            raise error_class.new("nope")
+  described_class::DO_NOT_RETRY_ERRORS.each do |error_class|
```
drewhoskins-stripe, line 81: I'd advise against this change because it assumes `described_class::DO_NOT_RETRY_ERRORS` is specified correctly and digs into implementation details.  I'd rather the test be "adversarial" and assert what correct behavior should be.
Also, AFAICT, it no longer tests the cases where we do want to retry.  It seems wrong not to test the other side.

lib/temporal/client/retryer.rb
```diff
@@ -9,19 +9,17 @@ module Retryer
...
-          GRPC::FailedPrecondition,
-          GRPC::InvalidArgument,
-          # If the activity has timed out, the server will return this and will never accept a retry
-          GRPC::NotFound,
-          GRPC::PermissionDenied,
-          GRPC::Unauthenticated,
-          GRPC::Unimplemented,
-        ]
-      end
+      DO_NOT_RETRY_ERRORS = [
```
drewhoskins-stripe, line 12: 👍

spec/unit/lib/temporal/client/retryer_spec.rb
antstorm: @drewhoskins-stripe do you have a failure mode in mind? Would it be someone adding an error to the list that should be retryable instead? But there isn't a great way to test against that unless you run test all possible error classes, which is not a real solution.

The retry case is covered by the spec above this one, but I see now that it's a bit different, I'll add the missing bit.

antstorm: Actually looking at this again… the retry case is identical to the spec above because of the `if !retried` escape hatch — in the case when a retry is expected the error wasn't getting raised outside of the `.with_retries` block, so therefore not picked up by rescue and not checked. This is however covered by a few specs above, e.g. `"backs off and stops retrying eventually"`

drewhoskins-stripe: These are not the same test.  In one case, you retry a few times.  In this one, you don't retry at all.

drewhoskins-stripe: > do you have a failure mode in mind?

Yep, it's a bit subtle but worth discussing.
Tests are the documentation of what the code is expected to do.  They represent the "contract."
If a test of the public-facing API changes, a code reviewer should scrutinize it more.
Suppose, the retryer is used in several locations.  Perhaps an author removes one of these errors because, for their place where the retryer is being used, we should retry.
A test should fail to alert the author (and code reviewer) that a key contract is being changed, and can go look at where the retryer is used.  They can then have the conversation of whether the retryer should have a dynamic list of errors that's passed in.

antstorm: > These are not the same test. In one case, you retry a few times. In this one, you don't retry at all.
I don't get it. In the current version you would retry once and then throw, which is no different from say `"can succeed after retries"` test case or `"executes the on_retry callback"` if the intention is to count callback calls. What exact use-case do you think is not covered after the proposed change?

I agree that having a "contract" in place is beneficial, it practise what would happen is whoever removes an error from the list will also update the spec with an equal change. But there's another side to it — someone might add an error to the list and not update the spec. In this case the spec will not fail and the contract gets broken. That can probably be solved by a specific test that checks which exact errors are listed as non-retriable

PR #89, opened 2021-07-13
Fix float serialization roundtrip

drewhoskins-stripe:
# Description
Timestamps I was using for high-precision cursoring were getting truncated.
This approach gets the correct answer, the disadvantage being that when the numbers get weird enough, it switches to scientific notation.

e.g. `1626122511.3059866` becomes `0.16261225113059866e10`
Certainly an improvement over `1626122511.30599` which was what happened before.

# Test Plan
`bundle exec rspec spec/unit/lib/temporal/client/converter/payload/json_spec.rb`

### DIFF
--- a/lib/temporal/json.rb
+++ b/lib/temporal/json.rb
@@ -4,7 +4,9 @@ require 'oj'
 module Temporal
   module JSON
     OJ_OPTIONS = {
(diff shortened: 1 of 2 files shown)


antstorm (approved):
👍

PR #92, opened 2021-08-30
[Refactor] Use local configuration instead of a global one

DeRauk:
This is a sync of https://github.com/coinbase/cadence-ruby/pull/41

To quote @anthonydmitriyev's  description:

_I'm working on a bigger change that will turn ~~Cadence~~ Temporal into an initializable class rather than using a single global instance. This will allow connecting your app to multiple clusters (this is handy for easy migrating from one ~~Cadence~~ Temporal cluster to another or more exotic solutions such as client-side sharding)._

_This is the first step which makes configuration local, passed to all classes that are using it_


antstorm (approved):
Looks good, thank you!

PR #97, opened 2021-09-17
Implement strategies for resetting workflows

DeRauk:
Syncing [#48](https://github.com/coinbase/cadence-ruby/pull/48).

Original PR Description:
"
This change would allow specifying a strategy when resetting a workflow. This is very useful when resetting a batch of workflows that all might have failed at different points in history

~~Tested against a running server & specs updated to reflect the change~~
"

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ _yardoc/
 
 # rspec failure tracking
 .rspec_status
+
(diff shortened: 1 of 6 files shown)


antstorm (approved):
Brilliant, thank you @DeRauk!

PR #100, opened 2021-09-28
Support signaling workflows in a namespace

drewhoskins-stripe:
# Description
Allow a namespace to be specified when signalling a workflow.

# Test Plan
There was no CI test coverage of signal_workflow, so we added a basic test as well as one with a namespace.
`rspec ./spec/unit/lib/temporal/client_spec.rb`


antstorm (approved):
Nice one 👍

PR #101, opened 2021-09-28
Handle server timeouts in await_workflow_result

drewhoskins-stripe:
# Description
Occasionally when leaving the timeout at thirty seconds on the client, we would see an error trying to access `type` on `nil` rather than the expected `TimeoutError`
This is because the server can give up first and return an empty list.
This handles that case gracefully and gives a better/more actionable error

# Test plan
`rspec ./spec/unit/lib/temporal/client_spec.rb`
integration tests:
`rspec ./spec/integration/await_workflow_result_spec.rb`

### DIFF
--- a/lib/temporal/client.rb
+++ b/lib/temporal/client.rb
@@ -104,6 +104,15 @@ module Temporal
           timeout: timeout || max_timeout,
         )
       rescue GRPC::DeadlineExceeded => e
+        # client timed out
+        closed_event = nil
+      else
+        history = Workflow::History.new(history_response.history.events)
+        closed_event = history.events.first
+        # If this is nil, the server timed out
+      end
+
+      if closed_event.nil?
         message = if timeout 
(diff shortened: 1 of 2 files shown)


antstorm (commented):
The logic here makes perfect sense to me, but I wonder if we should be handling this case on the client and assuming a server timeout. I merge [a similar change](https://github.com/coinbase/temporal-ruby/pull/104) for the `GetWorkflowExecutionHistory`, but I'm classifying it as an `UnexpectedResponse` since the message here is clearly malformed and it's a server-side issue. I wonder what are your thoughts on this?

drewhoskins-stripe:
In this case, I believe the behavior is intentional.  Both the integration test service and the production service behave in this way, returning an empty list after 30 seconds.

drewhoskins-stripe:
Hmm, on further testing, this is consistently happening after 20 seconds.

antstorm (approved):
Does it mean your theory isn't correct? Or something else?

PR #103, opened 2021-09-28
Gracefully degrade application failure error when deserializing

drewhoskins-stripe:
# Description
As discussed in [slack](https://temporalio.slack.com/archives/C01DQSMS8GM/p1631923998007500), our error handler raises when confronted with a user error that takes multiple arguments to instantiate.

This improves it by at least raising an error that links back to the original error (and still keeping the original backtrace)

Also log an error when this happens.

# Test Plan
Unit:
`rspec ./spec/unit/lib/temporal/workflow/errors_spec.rb`
Integration, to check for regressions:

```
cd examples
./bin/worker
rspec ./spec/integration/await_workflow_result_spec.rb
```


spec/fabricators/application_failure_fabricator.rb
```diff
@@ -0,0 +1,12 @@
+# Simulates Temporal::Connection::Serializer::Failure
+Fabricator(:application_failure, from: Temporal::Api::Failure::V1::Failure) do
```
antstorm, line 2: nit: Can you please move this one over to `fabricators/grpc` and prefix with `api_`? Helps with distinguishing between domain and edge entities

spec/fabricators/application_failure_fabricator.rb
```diff
@@ -0,0 +1,12 @@
...
+  transient :error_class, :backtrace
+  message { |attrs| attrs[:message] }
+  stack_trace { |attrs| attrs[:backtrace].join("\n") }
+  application_failure_info do |attrs|
+    Temporal::Api::Failure::V1::ApplicationFailureInfo.new(
+      type: attrs[:error_class],
+      details: TestDeserializer.new.to_details_payloads(attrs[:message]),
```
antstorm, line 9: This will fail when used in any other spec since `TestDeserializer` is only defined in a your new specs

antstorm: Also since the fabricator is for a proto message it probably should not be concerned with error serialisation, but passed details for the error directly

spec/unit/lib/temporal/workflow/errors_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      expect(e).to be_a(StandardError)
+      expect(e.message).to eq("ErrorWithTwoArgs: An error message")
+      expect(e.backtrace).to eq(stack_trace)
+      expect(Temporal.logger)
+        .to have_received(:error)
+        .with(
+          'Could not instantiate original error. Defaulting to StandardError.',
+          {
+            :original_error=>"ErrorWithTwoArgs",
+            :instantiation_error_message=> "wrong number of arguments (given 1, expected 2)",
```
antstorm, line 80: Any specific reason for not using the new Hash syntax here (`key: 'value'`)?

antstorm (changes requested):

spec/fabricators/application_failure_fabricator.rb
drewhoskins-stripe: Good flag.  As for the second suggestion, feels like reusing production methods rather than reimplementing them trumps strict dependencies here.

antstorm (approved):
Looks good, thank you for the PR, Drew. I've noticed that it's missing a test case that uses an explicitly provided default error class.

It might be really useful to provide one of our error classes based on the context instead of the generic `StandardError` (which might be difficult to distinguish from other failures).

Let's add this in a follow up PR, but we should do this before bumping the version

PR #105, opened 2021-09-30
Add ParentClosePolicy

tim37021:
I found [`ParentClosePolicy`](https://docs.temporal.io/docs/content/what-is-a-parent-close-policy/) is not supported. Here's the PR.

## Story
I am implementing proxy workflow pattern mention [HERE](https://community.temporal.io/t/sync-wait-a-workflow/3033)
I want to launch a child workflow and the main workflow will finish earlier then it's child. Due to default parent close policy. The child workflow will be killed as well. 

### without `parent_close_policy`
image

### with `parent_close_policy: :abandon`
image

## Usage
```ruby
class ProxyWorkflow < Temporal::Workflow
  def execute(workflow_id)
    # copied from SDK example
    future = LongRunningActivity.execute(10, 1)
    workflow.execute_workflow(
      MainWorkflow,
      proxy_workflow_id: workflow_id,
      proxy_run_id: workflow.metadata.run_id,
      options: {
        parent_close_policy: :abandon,
      },
    )
    ret = nil
    workflow.on_signal do |signal, input|
      future.cancel
      ret = input
    end
    future.wait
    ret
  end
end
```

```ruby
class MainWorkflow < Temporal::Workflow
  def execute(proxy_workflow_id:, proxy_run_id:)
    workflow.sleep(5)
    SignalActivity.execute!(proxy_workflow_id: proxy_workflow_id, proxy_run_id: proxy_run_id, input: "succsss yeah")
    10.times do
      HelloActivity.execute!("hehehe")
      workflow.sleep(1)
    end
  end
end
```

By the way, how can I know my workflow id? Currently I explicitly names it.
```ruby
Temporal.start_workflow(TemporalWorkflow::ProxyWorkflow,  "hehehe", options: {workflow_id: "hehehe"})
```

### DIFF
--- a/lib/temporal/execution_options.rb
+++ b/lib/temporal/execution_options.rb
@@ -1,5 +1,6 @@
 require 'temporal/concerns/executable'
 require 'temporal/retry_policy'
+require 'temporal/parent_close_policy'
 
 module Temporal
   class ExecutionOptions
--- a/lib/temporal/execution_options.rb
+++ b/lib/temporal/execution_options.rb
@@ -3,7 +3,7 @@ require 'temporal/retry_policy'
 
 module Temporal
   class ExecutionOptions
-    attr_reader :name, :namespace, :task_queue, :retry_policy, :timeouts, :headers
+    attr_reader :name, :namespace, :task_queue, :retry_policy, :parent_close_policy, :timeouts, :headers
 
     def initialize(object, options, defaults = nil)
       # Options are treated as overrides and take precedence
@@ -11,6 +11,7 @@ module Temporal
       @namespace = options[:namespace]
       @task_queue = options[:task_queue] || options[:task_list]
       @retry_policy = options[:retry_policy] || {}
+      @parent_close_policy = options[:parent_close_policy] || {}
       @timeouts = options[:timeouts] || {}
       @headers = options[:headers] || {}
 
@@ -36,6 +37,13 @@ module Temporal
       else
         @retry_policy = Temporal::RetryPolicy.new(@retry_policy)
         @retry_policy.validate!
+      end
+
+      if @parent_close_policy.empty?
+        @parent_close_policy = nil
+      else
+        @parent_close_policy = Temporal::ParentClosePolicy.new(@parent_close_policy)
+        @parent_close_policy.validate!
       end
 
       freeze
--- a/lib/temporal/parent_close_policy.rb
+++ b/lib/temporal/parent_close_policy.rb
@@ -1,0 +1,20 @@
+require 'temporal/errors'
+
+module Temporal
+  # See https://docs.temporal.io/docs/content/what-is-a-parent-close-policy/ for more information
+  class ParentClosePolicy
+    class InvalidParentClosePolicy < ClientError; end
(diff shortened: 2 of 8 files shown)


lib/temporal/parent_close_policy.rb
```diff
@@ -0,0 +1,20 @@
...
+    class InvalidParentClosePolicy < ClientError; end
+
+    attr_reader :policy
+
+    def initialize(policy)
+      @policy = policy
+    end
+
+    def validate!
+      unless %i[terminate abandon request_cancel].include?(@policy)
```
antstorm, line 15: Can you please define these as constants so they can be used without hardcoding symbols?

lib/temporal/execution_options.rb
```diff
@@ -1,16 +1,18 @@
...
-    attr_reader :name, :namespace, :task_queue, :retry_policy, :timeouts, :headers
+    attr_reader :name, :namespace, :task_queue, :retry_policy, :parent_close_policy, :timeouts, :headers
 
     def initialize(object, options, defaults = nil)
       # Options are treated as overrides and take precedence
       @name = options[:name] || object.to_s
       @namespace = options[:namespace]
       @task_queue = options[:task_queue] || options[:task_list]
       @retry_policy = options[:retry_policy] || {}
+      @parent_close_policy = options[:parent_close_policy] || {}
```
antstorm, line 15: ExecutionOptions are shared between workflows and activities, so this wouldn't be the best place to add workflow-specific options. Can you please move these inline with the child workflow invocation?

spec/unit/lib/temporal/connection/serializer/parent_close_policy_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      end.not_to raise_error
+
+      expect do
+        described_class.new(abandon_policy).to_proto
+      end.not_to raise_error
+
+      expect do
+        described_class.new(request_cancel_policy).to_proto
+      end.not_to raise_error
+    end
```
antstorm, line 22: Can you please split these into 3 examples? It's easier to figure out which ones have failed that way

tim37021:
> Good call on getting the workflow_id when starting a workflow without explicitly providing it, we don't have a way to do that currently. 
I think it's possible to not to have break change? instead of changing the interface.

I remember we have a `run_id` inside `workflow.metadata`
but it doesn't contains `workflow_id` currently. is it good to add that inside metadata?

antstorm:
@tim37021 yes, I was thinking that the most logical solution would be to return both workflow_id and run_id for a `start_workflow` call:

```ruby
workflow_id, run_id = Temporal.start_workflow(MyWorkflow)
```

However exposing this via metadata still works, but I'm not sure how you're planning to get the metadata.

As for adding workflow_id to metadata — yes, that's something I've already implemented for cadence — https://github.com/coinbase/cadence-ruby/pull/60, I'll replay that change over this repo today since others have asked about it as well

antstorm:
There we go — https://github.com/coinbase/temporal-ruby/pull/110

PR #106, opened 2021-10-01
Timer operations testing sync

DeRauk:
https://github.com/coinbase/cadence-ruby/pull/57

### DIFF
--- a/lib/temporal/testing/temporal_override.rb
+++ b/lib/temporal/testing/temporal_override.rb
@@ -1,4 +1,5 @@
 require 'securerandom'
+require 'temporal/thread_local_context'
 require 'temporal/activity/async_token'
 require 'temporal/workflow/execution_info'
 require 'temporal/testing/workflow_execution'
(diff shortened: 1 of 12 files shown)


lib/temporal/testing/temporal_override.rb
```diff
@@ -95,6 +107,8 @@ def start_locally(workflow, schedule, *input, **args)
           execution, workflow_id, run_id, workflow.disabled_releases, metadata
         )
 
+        Temporal::ThreadLocalContext.set(context)
```
DeRauk, line 110: Two tests are failing and I have a hunch it's related to this line. 
 - `./spec/unit/lib/temporal/testing/temporal_override_spec.rb:106 # Temporal::Testing::TemporalOverride when testing mode is local Workflow.execute_locally restores original context after finishing successfully`
 - `/spec/unit/lib/temporal/testing/temporal_override_spec.rb:117 # Temporal::Testing::TemporalOverride when testing mode is local Workflow.execute_locally restores original context after failing`
 
 They don't fail when run by themselves, just when the whole test suite is run. Not sure what the best way to fix is.
 
 These tests aren't in the Cadence library.

PR #107, opened 2021-10-05
Pass config to the error handler

DeRauk:
Right now the error handler is using the global config, we want it to use the config that the client was created with.

### DIFF
--- a/lib/temporal/activity/poller.rb
+++ b/lib/temporal/activity/poller.rb
@@ -78,7 +78,7 @@ module Temporal
       rescue StandardError => error
         Temporal.logger.error("Unable to poll activity task queue", { namespace: namespace, task_queue: task_queue, error: error.inspect })
 
-        Temporal::ErrorHandler.handle(error)
+        Temporal::ErrorHandler.handle(error, config)
 
         nil
       end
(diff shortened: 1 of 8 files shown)


lib/temporal/activity/poller.rb
```diff
@@ -78,7 +78,7 @@ def poll_for_task
       rescue StandardError => error
         Temporal.logger.error("Unable to poll activity task queue", { namespace: namespace, task_queue: task_queue, error: error.inspect })
 
-        Temporal::ErrorHandler.handle(error)
+        Temporal::ErrorHandler.handle(error, config)
```
drewhoskins-stripe, line 81: I think you'll find this is the wrong config.  See line 48 in this file.

lib/temporal/workflow.rb
```diff
@@ -20,7 +20,7 @@ def self.execute_in_context(context, input)
       Temporal.logger.error("Workflow execution failed", context.metadata.to_h.merge(error: error.inspect))
       Temporal.logger.debug(error.backtrace.join("\n"))
 
-      Temporal::ErrorHandler.handle(error, metadata: context.metadata)
+      Temporal::ErrorHandler.handle(error, context.config, metadata: context.metadata)
```
DeRauk, line 23: This line is what's breaking the tests, need to dig in a little more to find why.

antstorm (approved):
Looks good!

PR #108, opened 2021-10-15
Pull workflow_id and namespace into Workflow metadata for better logging

drewhoskins-stripe:
# Description
Unlike activities, I noticed our workflow logging didn't have access to the workflow_id or namespace, nor did the underlying context.
This will give user-defined loggers access so they can log consistency for the entire duration of a workflow.

Also
* I'm renaming workflow_run_id to run_id in the logs, as that's the canonical term used everywhere and can be used for log filter queries.  The only place it's slightly weird is in activities, but they do not have their own concept of run_id, so it should be unambiguous.
* Implementation detail: I removed a layer of indirection for creating metadata that was getting in the way.

# Test Plan
Some unit tests:
```
bundle exec rspec spec/unit/lib/temporal/metadata/workflow_spec.rb
bundle exec rspec ./spec/unit/lib/temporal/metadata_spec.rb
```
Plus the ambient integration tests
```
cd examples
# Terminal 1
./bin/worker
# Terminal 2
USE_ENCRYPTION = 1 ./bin/worker
# Terminal 3
bundle exec rspec spec/integration/
```
And manual inspection of logging.

### DIFF
--- a/lib/temporal/metadata.rb
+++ b/lib/temporal/metadata.rb
@@ -6,37 +6,11 @@ require 'temporal/concerns/payloads'
 
 module Temporal
   module Metadata
-    ACTIVITY_TYPE = :activity
-    WORKFLOW_TASK_TYPE = :workflow_task
-    WORKFLOW_TYPE = :workflow
 
     class << self
       include Concerns::Payloads
 
-      def generate(type, data, namespace = nil)
-        case type
-        when ACTIVITY_TYPE
-          activity_metadata_from(data, namespace)
-        when WORKFLOW_TASK_TYPE
-          workflow_task_metadata_from(data, namespace)
-        when WORKFLOW_TYPE
-          workflow_metadata_from(data)
-        else
-          raise InternalError, 'Unsupported metadata type'
-        end
-      end
-
-      private
-
-      def headers(fields)
-        result = {}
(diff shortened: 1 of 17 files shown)


lib/temporal/metadata.rb
```diff
@@ -6,37 +6,11 @@
...
 module Temporal
   module Metadata
-    ACTIVITY_TYPE = :activity
-    WORKFLOW_TASK_TYPE = :workflow_task
-    WORKFLOW_TYPE = :workflow
 
     class << self
       include Concerns::Payloads
 
-      def generate(type, data, namespace = nil)
```
drewhoskins-stripe, line 16: I removed this layer of indirection because we need to pass in different arguments depending on the type of metadata.

antstorm (approved):
I've noticed that this logic has already been implemented in Cadence SDK (https://github.com/coinbase/cadence-ruby/pull/60) and not synced with this repo. Can you please do that? I'm happy to do it myself as well, just trying to keep these repos directionally similar to make sure bigger changes can also be replicated without too much extra work

antstorm (changes requested):
Sorry, meant to request changes, not approve

drewhoskins-stripe:
@jeffschoner-stripe is going to handle this one.

PR #110, opened 2021-10-25
[Feature] Add id and domain to workflow context's metadata

DeRauk:
This is syncing https://github.com/coinbase/cadence-ruby/pull/60

Original PR description:
```
There are 3 types of metadata — decision, activity and workflow. Each is exposed to different parts of the SDK.

Workflow ID is generated from the history, which confusingly doesn't have domain or workflow_id. This PR adds these 2 missing fields from the decision metadata

Tested using unit specs and manual workflow runs
```


antstorm (approved):
Perfect, thank you, @DeRauk!

AbulKhayer2244 (approved):

PR #112, opened 2021-11-01
Add signal arguments to start_workflow (to support signal_with_start)

nagl-stripe:
This PR adds support for signal_with_start_workflow_execution. It is a rework of https://github.com/coinbase/temporal-ruby/pull/99 based on Anthony's feedback.

New tests:

bundle exec rspec spec/unit/lib/temporal_spec.rb
bundle exec rspec spec/unit/lib/temporal/client_spec.rb
bundle exec rspec spec/unit/lib/temporal/grpc_client_spec.rb
cd examples && bundle exec rspec spec/integration/signal_with_start_spec.rb


antstorm (changes requested):
This is great, thank you for addressing all the comments!

One small thing before we merge this — can we make `signal_name` and `signal_input` to be part of the `options` that are passed to `start_workflow` rather than top-level arguments? This makes it easier to distinguish workflow-specific arguments from SDK-specific options

lib/temporal/testing/temporal_override.rb
```diff
@@ -73,6 +73,11 @@ def start_locally(workflow, schedule, *input, **args)
         options = args.delete(:options) || {}
         input << args unless args.empty?
 
+        # signals aren't supported at all, so let's prohibit start_workflow calls that try to signal
+        signal_name = options.delete(:signal_name)
+        signal_input = options.delete(:signal_input)
+        raise NotImplementedError, 'Signals are not available when Temporal::Testing.local! is on' unless signal_name.nil? && signal_input.nil?
```
antstorm, line 79: nit: inverted condition would probably read much better here — `if signal_name || signal_input`

lib/temporal/client.rb
```diff
@@ -13,26 +13,60 @@ def initialize(config)
       @config = config
     end
 
+    # Starts a workflow with an optional signal.
```
antstorm, line 16: Can you please merge this with the `master` branch? I've added YARD-style documentation that should give you some structure for your comments

antstorm (approved):
Brilliant, thank you for your contribution! 🙌

PR #113, opened 2021-11-03
Add YARD documentation for Temporal::Client

antstorm:
Adds a YARD-style documentation for the project which can be viewed over at https://rubydoc.info/github/coinbase/temporal-ruby

### DIFF
--- a/lib/temporal/client.rb
+++ b/lib/temporal/client.rb
@@ -13,6 +13,23 @@ module Temporal
       @config = config
     end
 
+    # Start a workflow
+    #
+    # @param workflow [Temporal::Workflow, String] workflow class or name. When a workflow class
+    #   is passed, its config (namespace, task_queue, timeouts, etc) will be used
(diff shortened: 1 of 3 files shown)


lib/temporal/client.rb
```diff
@@ -13,6 +13,23 @@ def initialize(config)
       @config = config
     end
 
+    # Start a workflow
+    #
+    # @param workflow [Temporal::Workflow, String] workflow class or name. When a workflow class
+    #   is passed, its config (namespace, task_queue, timeouts, etc) will be used
+    # @param input [any] arguments to be passed to workflow's #execute method
+    # @param args [Hash] keyword arguments to be pass to workflow's #execute method
```
DeRauk, line 21: spelling nit: to be `passed` to ...

DeRauk (approved):

lib/temporal/client.rb
antstorm: Nice catch! Fixed

PR #119, opened 2021-11-16
Turn off schedule_to_start activity timeout by default

drewhoskins-stripe:
# Description
As shown in the comments, the temporal team strongly recommends against using `schedule_to_start` timeout unless you're doing things like manually re-routing to different task queues.
We have also run into struggles with this timeout.  I think the best policy is to remove the default.

# Test Plan
`bundle exec rspec ./spec/unit/lib/temporal/configuration_spec.rb`


antstorm (approved):
I think that works and even though it's breaking I don't think anyone actually depends on this timeout, so happy to proceed 👍

PR #122, opened 2021-11-20
Add describe_namespace

drewhoskins-stripe:
# Description

Exposing `describe_namespace` GRPC in our namespace

# Test Plan
Integration test:
Terminal 1:
`stripe/temporal-ruby/examples $ ./bin/worker `
Terminal 2:
`rspec spec/integration/describe_namespace_spec.rb`

Unit test:
`rspec spec/unit/lib/temporal/client_spec.rb`


spec/unit/lib/temporal/client_spec.rb
```diff
@@ -310,6 +310,18 @@ def expect_signal_with_start(expected_arguments, expected_signal_argument)
...
+  describe '#describe_namespace' do
+    before { allow(connection).to receive(:describe_namespace).and_return(Temporal::Api::WorkflowService::V1::DescribeNamespaceResponse.new) }
+    
+    it 'passes the namespace to the connection' do
+      result = subject.describe_namespace('new-namespace')
+
+      expect(connection)
+        .to have_received(:describe_namespace)
+        .with(name: 'new-namespace')
+      expect(result).to be_an_instance_of(Temporal::Api::WorkflowService::V1::DescribeNamespaceResponse)
```
antstorm, line 322: nit: This assertion doesn't add any value since it's only making sure Spec works as expected

lib/temporal/client.rb
```diff
@@ -133,6 +133,10 @@ def register_namespace(name, description = nil)
       connection.register_namespace(name: name, description: description)
     end
 
+    def describe_namespace(name)
```
antstorm, line 136: Can you please add a YARD comment similar to other methods in this file? It's the widest public-facing API, so would be good to keep it documented

examples/spec/integration/describe_namespace_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'temporal/errors'
+
+describe 'Temporal.describe_namespace' do
+  it 'returns a value' do
+    description = 'Namespace for temporal-ruby integration test'
+    begin
+      Temporal.register_namespace('a_test_namespace', description)
+    rescue Temporal::NamespaceAlreadyExistsFailure
+    end
+    result = Temporal.describe_namespace('a_test_namespace')
```
antstorm, line 10: Does this work every time without flakes? I had to `sleep 15` during the integration spec setup to allow it to settle

antstorm (commented):
Thank you! Left a few minor comments

examples/spec/integration/describe_namespace_spec.rb
drewhoskins-stripe: I have never seen it flake and we've been running it for a couple months.

antstorm (approved):
LGTM, thanks!

PR #125, opened 2021-12-07
[Fix] Non-started activity cancellation

antstorm:
When an activity cancellation is requested before the activity had a chance to start Temporal will issue an `ACTIVITY_CANCELED` event. This PR fixed handling of this event.

Tested:
- Added both unit & integration specs to ensure it works as expected

### DIFF
--- a/examples/spec/integration/activity_cancellation_spec.rb
+++ b/examples/spec/integration/activity_cancellation_spec.rb
@@ -1,0 +1,39 @@
+require 'workflows/long_workflow'
+
+describe 'Activity cancellation' do
+  let(:workflow_id) { SecureRandom.uuid }
+
+  it 'cancels a running activity' do
+    run_id = Temporal.start_workflow(LongWorkflow, options: { workflow_id: workflow_id })
+
+    # Signal workflow after starting, allowing it to schedule the first activity
+    sleep 0.5
+    Temporal.signal_workflow(LongWorkflow, :CANCEL, workflow_id, run_id)
+
+    result = Temporal.await_workflow_result(
+      LongWorkflow,
+      workflow_id: workflow_id,
+      run_id: run_id,
+    )
+
+    expect(result).to be_a(LongRunningActivity::Canceled)
+    expect(result.message).to eq('cancel activity request received')
+  end
+
+  it 'cancels a non-started activity' do
+    # Workflow is started with a signal which will cancel an activity before it has started
+    run_id = Temporal.start_workflow(LongWorkflow, options: {
+      workflow_id: workflow_id,
+      signal_name: :CANCEL
+    })
+
+    result = Temporal.await_workflow_result(
+      LongWorkflow,
+      workflow_id: workflow_id,
+      run_id: run_id,
+    )
+
+    expect(result).to be_a(Temporal::ActivityCanceled)
+    expect(result.message).to eq('ACTIVITY_ID_NOT_STARTED')
+  end
+end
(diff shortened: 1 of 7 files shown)


lib/temporal/errors.rb
```diff
@@ -19,6 +19,9 @@ class TimeoutError < ClientError; end
   # with the intent to propagate to a workflow
   class ActivityException < ClientError; end
 
+  # Represents cancellation of a non-started activity
```
drewhoskins-stripe, line 22: Since cancels can also happen at a heartbeat, should we unify this case with that case?  It's not ideal that people would have to deal with both depending on the timing of when the cancellation comes in.
If that's not possible, should name it as `ActivityCanceledBeforeStart`?

spec/unit/lib/temporal/workflow/history/event_target_spec.rb
```diff
@@ -21,5 +21,21 @@
         expect(subject.type).to eq(described_class::CANCEL_TIMER_REQUEST_TYPE)
       end
     end
+
+    context 'when event is ACTIVITY_CANCELED' do
+      let(:raw_event) { Fabricate(:api_activity_task_canceled_event) }
+
+      it 'sets type to timer' do
```
drewhoskins-stripe, line 28: copypasta?

spec/unit/lib/temporal/workflow/history/event_target_spec.rb
```diff
@@ -21,5 +21,21 @@
...
+        expect(subject.type).to eq(described_class::ACTIVITY_TYPE)
+      end
+    end
+
+    context 'when event is ACTIVITY_TASK_CANCEL_REQUESTED' do
+      let(:raw_event) { Fabricate(:api_activity_task_cancel_requested_event) }
+
+      it 'sets type to cancel_timer_request' do
```
drewhoskins-stripe, line 36: ©️ 🍝 here as well

lib/temporal/errors.rb
antstorm: There's nothing preventing other to re-use the `Temporal::ActivityCancelled` error for heartbeat cancellations. It's just that a heartbeat cancellation is no different from any other failure raised by an activity. I think a more interesting question here is — should we provide a `#cancel!` API for the activity context that would raise that error?

drewhoskins-stripe: Oh, I forgot that in this SDK, heartbeat cancel doesn't throw, it returns a boolean.
In our wrapper SDK (and in the java SDK), it will throw.  We think this is the best way to make sure developers don't drop cancellations.

examples/spec/integration/activity_cancellation_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  let(:workflow_id) { SecureRandom.uuid }
+
+  it 'cancels a running activity' do
+    run_id = Temporal.start_workflow(LongWorkflow, options: { workflow_id: workflow_id })
+
+    # Signal workflow after starting, allowing it to schedule the first activity
+    sleep 0.5
+    Temporal.signal_workflow(LongWorkflow, :CANCEL, workflow_id, run_id)
+
+    result = Temporal.await_workflow_result(
```
drewhoskins-stripe, line 13: This is always failing for me, locally
```
  1) Activity cancellation cancels a running activity
     Failure/Error: expect(result).to be_a(LongRunningActivity::Canceled)
       expected #<Temporal::ActivityCanceled: ACTIVITY_ID_NOT_STARTED> to be a kind of LongRunningActivity::Canceled
     # ./spec/integration/activity_cancellation_spec.rb:19:in `block (2 levels) in <top (required)>'
```

PR #132, opened 2022-01-19
Add namespace to emitted metrics

christopherb-stripe:
This PR adds the namespace as a tag to the metrics that were missing it. The tag name was chosen to be consistent with the  other times it is used with other metrics in the library. This is helpful to be able to tell if a particular namespace is having issues. The current activity/workflow name tag doesn't have enough context.


DeRauk:
Looks good, thanks!

DeRauk (approved):

PR #133, opened 2022-02-01
Add workflow tag to metrics emitted from the activity task processor

christopherb-stripe:
We've found it helpful to tag activity metrics with the workflow name that called them so that we can build a dashboard that includes all related information for a workflow. The tag is named the same as other workflow tags in the library e.g. in the workflow task processor.

This PR also sneaks a fix for a logging typo in too.


DeRauk:
Thanks @christopherb-stripe!

PR #135, opened 2022-02-05
Validate commands issued after workflow completion

drewhoskins-stripe:
# Description

Customers have seen some difficult-to-diagnose errors from the server when they let activities dangle after the workflow ends.
Examples:
https://github.com/temporalio/temporal/blob/744f26d282c5891204d66aac0d77c93cdf1c1291/service/history/commandChecker.go#L789
https://github.com/temporalio/temporal/blob/43b4531277b4d5b519596ce4e78eaf778f572c99/service/history/workflowTaskHandlerCallbacks.go#L205

These errors are thrown when the server receives a terminal event followed by other events.  (I haven't figured out what differentiates which error is hit.)
Typically this seems to happen when teams don't wait for all their async activities before the workflow ends.
This error should be clearer and give them a proper callstack.
I made it a task failure so that the workflow will be retried and they can deploy to fix.

# Test plan
```
temporal-ruby $ rspec -f d ./spec/unit/lib/temporal/workflow/state_manager_spec.rb 
[Coveralls] Set up the SimpleCov formatter.
[Coveralls] Using SimpleCov's default settings.

Temporal::Workflow::StateManager
  #schedule
    fails to validate if Temporal::Workflow::Command::ContinueAsNew is not the last command scheduled
    fails to validate if Temporal::Workflow::Command::FailWorkflow is not the last command scheduled
   fails to validate if Temporal::Workflow::Command::CompleteWorkflow is not the last command scheduled

Finished in 0.00378 seconds (files took 1.04 seconds to load)
3 examples, 0 failures
```

`./bin/trigger InvalidContinueAsNewWorkflow`

Output:
```
E, [2022-02-04T17:53:57.107667 #57509] ERROR -- temporal_client: Workflow execution failed {"namespace":"ruby-samples","workflow_id":"2728fcb7-16ba-4fcd-b9c1-9119f72e7287","workflow_name":"InvalidContinueAsNewWorkflow","workflow_run_id":"db086762-c5a5-49f1-8334-bf477fed6b28","attempt":1,"task_queue":"general","run_started_at":1644026025.9409,"memo":{},"error":"#<Temporal::TryingToCompleteWorkflowError: You cannot do anything in a Workflow after it completes. You called continue_as_new. But then we received a new command: Temporal::Workflow::Command::ScheduleActivity.  This can happen, for example, if you've not waited for all of your Activity futures before finishing the Workflow.>"}
D, [2022-02-04T17:53:57.107744 #57509] DEBUG -- temporal_client: /Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow/state_manager.rb:110:in `block in validate_append_command'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow/state_manager.rb:99:in `each'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow/state_manager.rb:99:in `validate_append_command'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow/state_manager.rb:59:in `schedule'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow/context.rb:316:in `schedule_command'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow/context.rb:64:in `execute_activity'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/activity/workflow_convenience_methods.rb:23:in `execute'
/Users/drewhoskins/coinbase/temporal-ruby/examples/workflows/invalid_continue_as_new_workflow.rb:12:in `block in execute'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow/future.rb:59:in `done'
/Users/drewhoskins/coinbase/temporal-ruby/examples/workflows/invalid_continue_as_new_workflow.rb:11:in `execute'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow.rb:16:in `execute_in_context'
/Users/drewhoskins/coinbase/temporal-ruby/lib/temporal/workflow/executor.rb:48:in `block in execute_workflow' {}
```


lib/temporal/errors.rb
```diff
@@ -47,6 +47,10 @@ def initialize(new_run_id:)
     end
   end
 
+  # Thrown when we can't complete a workflow due to operations occurring after the workflow
+  # is finished.  Could be due to signals, dangling asynchronous activities, or internal framework bug.
+  class TryingToCompleteWorkflowError < InternalError; end
```
antstorm, line 52: I think this error name is a bit confusing, what about `WorkflowHasAlreadyStopped`? Or something similar

antstorm (changes requested):
I like the idea, thanks for raising this, @drewhoskins-stripe !

I however think this might be inefficient is some cases (when you are trying to achieve a high level of concurrency within a workflow). We already have a `completed?` flag on the workflow context, why not reuse it and raise the error from `Workflow::Context#schedule_command`? You can also probably loose some of the message details (reporting workflow status)

lib/temporal/workflow/state_manager.rb
```diff
@@ -94,6 +95,27 @@ def next_event_id
         @last_event_id += 1
       end
 
+      def validate_append_command(command)
+        return if commands.last.nil?
+        _, previous_command = commands.last
```
drewhoskins-stripe, line 100: Sorry, I was an idiot; there was no need to loop since we can maintain the list incrementally.

antstorm (approved):
Thanks, Drew!

Before we merge this, any thoughts on moving the validation into the Workflow context as I suggested in the previous comment?

drewhoskins-stripe:
Sorry, I assumed the performance issue was your main point, which I fixed.
Context feels like the wrong abstraction layer, as this is basically "invalid state transition" and the state manager is what keeps track of the commands.

PR #137, opened 2022-02-09
Added list_namespace to client and fixed gRPC request

calum-stripe:
`list_namespace` was broken because of this PR merged into the Protos repo. This fixes this and also exposes `list_namespaces` as a function on the client (we're wanting to use this functionality at Stripe)


antstorm (changes requested):
Hey @calum-stripe !

The change looks good, that you for contributing! A couple of things to consider:

1. Can you please add a unit test (in addition to the integration test)?
2. I wonder if it makes sense to allow the caller to specify a page size without the offset? Otherwise there's no way to get to results beyond your page size without increasing it
3. I wonder if we should start wrapping responses from gRPC in order to prevent proto exposure to the users of the SDK. I know that other methods might have the same issue, so it's optional as far as this PR is concerned

calum-stripe:
Hey @antstorm! 👋 
> 1. Can you please add a unit test (in addition to the integration test)?

Added :) 

> 2. I wonder if it makes sense to allow the caller to specify a page size without the offset? Otherwise there's no way to get to results beyond your page size without increasing it

Woops overlooked this... you're right! Added `next_page_token` to the function params

> 3. I wonder if we should start wrapping responses from gRPC in order to prevent proto exposure to the users of the SDK. I know that other methods might have the same issue, so it's optional as far as this PR is concerned

Interesting point, not sure how you track work like this that needs to be done but feel free to add it there!

lib/temporal/client.rb
```diff
@@ -143,6 +143,14 @@ def describe_namespace(name)
       connection.describe_namespace(name: name)
     end
 
+    # Fetches all the namespaces.
+    #
+    # @param page_size [Integer] number of namespace results to return per page.
+    # @param next_page_token [String] page token to advancing the pagination (provide this )
```
antstorm, line 149: Nit: this comment looks unfinished, maybe something like "a optional pagination token returned by a previous list_namespaces call"

antstorm (approved):
Looks great, thank you for making the final changes. One last nit and we're good to go

lib/temporal/client.rb
calum-stripe: Good catch, fixed! :) thanks @antstorm

antstorm (approved):
Brilliant, thank you!

PR #139, opened 2022-02-16
Populate namespace field to all workflow/activity task RPCs

cvanderschuere:
The somewhat recently added `namespace` field on workflow & activity task RPCs is critical for any use of an [Authorizer](https://docs.temporal.io/docs/server/security/#authorizer-plugin-interface).

Without this change, the flow is:
1.  Workflow tasks poll successful
2. [locally] Workflow task started
3. [locally] Workflow task finished
4. Workflow task completed -- `#<GRPC::PermissionDenied: 7:Request unauthorized.>`
5. Workflow task failed (failure is also not being reported to the server)

Related auth code:
https://github.com/temporalio/temporal/blob/3c427e6a2d5e7bcdda5868f379a26f6eb578871f/common/authorization/interceptor.go#L126-L137

### DIFF
--- a/lib/temporal/activity/task_processor.rb
+++ b/lib/temporal/activity/task_processor.rb
@@ -54,7 +54,7 @@ module Temporal
 
       attr_reader :task, :namespace, :task_token, :activity_name, :activity_class,
       :middleware_chain, :metadata, :config
-      
+
       def connection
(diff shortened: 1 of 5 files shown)


antstorm (approved):
Looks great, thank you!

PR #140, opened 2022-02-16
Adds SSL support

cvanderschuere:
Closes https://github.com/coinbase/temporal-ruby/issues/70

This PR is an update of https://github.com/coinbase/temporal-ruby/pull/73 to work after the client/connection split.

This change enables the use of [mTLS](https://docs.temporal.io/docs/server/security/#encryption-in-transit-with-mtls) for client<>server communication.


lib/temporal/connection.rb
```diff
@@ -10,12 +10,17 @@ def self.generate(configuration)
...
       port = configuration.port
+      credentials = configuration.credentials || :this_channel_is_insecure
 
       hostname = `hostname`
       thread_id = Thread.current.object_id
       identity = "#{thread_id}@#{hostname}"
 
-      connection_class.new(host, port, identity)
+      if configuration.type == :grpc
+        connection_class.new(host, port, identity, credentials)
```
antstorm, line 20: We can either loose the condition (since we only have one type of connection) or pass `credentials` as `options` Hash (allowing for other extensions in the future). Either way different signatures violate Liskov substitution principle (these are not technically subclasses, but you still want to maintain a consistent interface)

lib/temporal/connection.rb
```diff
@@ -10,12 +10,17 @@ def self.generate(configuration)
       connection_class = CLIENT_TYPES_MAP[configuration.type]
       host = configuration.host
       port = configuration.port
+      credentials = configuration.credentials || :this_channel_is_insecure
```
antstorm, line 13: nit: How about putting the default value into `Configuration` (along with other defaults)?

antstorm (changes requested):
Thank you for resurrecting this, @cvanderschuere ! Left few comments, but similar to the original PR this needs unit specs to test the new behaviour

cvanderschuere:
@antstorm mind taking another look?

jlemesh:
Hi, is it possible to finish and merge this? I'd be happy to help. @antstorm, any other changes needed here?

cvanderschuere:
Completed with https://github.com/coinbase/temporal-ruby/pull/186

PR #141, opened 2022-02-18
Support for invoking and processing queries

dwillett:
Adding support for [queries](https://docs.temporal.io/docs/concepts/what-is-a-query/). 

This PR leaves a few things unimplemented:
* A "catch-all" query handler syntax similar to signal handlers.
* Support for reserved query handler behaviors such as responding to `__stack_trace`, or support for populating the query UX on the web interface.


examples/workflows/query_workflow.rb
```diff
@@ -0,0 +1,22 @@
+class QueryWorkflow < Temporal::Workflow
+  def execute
+    state = "waiting for cancel"
+    last_signal_received = nil
+    workflow.on_query("state") do |input|
```
antstorm, line 5: nit: let's use consistent terminology here — either `input` or `args` (as in the GPRC interface)

lib/temporal/metadata.rb
```diff
@@ -33,12 +35,13 @@ def generate_workflow_task_metadata(task, namespace)
           attempt: task.attempt,
           workflow_run_id: task.workflow_execution.run_id,
           workflow_id: task.workflow_execution.workflow_id,
-          workflow_name: task.workflow_type.name
+          workflow_name: task.workflow_type.name,
+          query_type: task.query&.query_type,
+          query_args: from_query_payloads(task.query&.query_args)
```
antstorm, line 40: I wonder what's the use-case for exposing query via metadata?

antstorm: Also if we do this we'll need to think through the case when `queries` is used and there are multiple queries

lib/temporal/metadata/workflow_task.rb
```diff
@@ -3,16 +3,18 @@
...
+      def initialize(namespace:, id:, task_token:, attempt:, workflow_run_id:, workflow_id:, workflow_name:, query_type:, query_args:)
         @namespace = namespace
         @id = id
         @task_token = task_token
         @attempt = attempt
         @workflow_run_id = workflow_run_id
         @workflow_id = workflow_id
         @workflow_name = workflow_name
+        @query_type = query_type
+        @query_args = query_args
```
antstorm, line 17: This smells like a "data clamp" and might benefit from our own `Query` struct combining the two together

lib/temporal/workflow/dispatcher.rb
```diff
@@ -6,21 +6,33 @@ class Dispatcher
 
       def initialize
         @handlers = Hash.new { |hash, key| hash[key] = [] }
+        @query_handlers = Hash.new { |hash, key| hash[key] = {} }
```
antstorm, line 9: I would recommend keeping the `Dispatcher` unaware of specific use-cases (such as queries). It's a general event dispatcher and there's really no difference between queries and other events. We can use the `target` to identify a specific query

lib/temporal/workflow/dispatcher.rb
```diff
@@ -6,21 +6,33 @@ class Dispatcher
...
+      end
+
       def dispatch(target, event_name, args = nil)
         handlers_for(target, event_name).each do |handler|
           handler.call(*args)
         end
       end
 
+      def process_query(target, query, args)
+        if query == '__cadence_web_list'
```
antstorm, line 27: This is also slightly misplaced making the `Dispatcher` even more aware of the high-level logic. This part should probably be handled by the `Workflow::Context`

lib/temporal/workflow/context.rb
```diff
@@ -287,6 +287,14 @@ def on_signal(&block)
         end
       end
 
+      def on_query(query, &block)
```
antstorm, line 290: I wonder if we should allow a "catch-all" query handler via `workflow.on_query do |query, input|` (without specifying a query). What do you think?

lib/temporal/workflow/task_processor.rb
```diff
@@ -38,7 +38,37 @@ def process
...
+          # } else {
+          # 	queryCompletedRequest.CompletedType = enumspb.QUERY_RESULT_TYPE_ANSWERED
+          # 	queryCompletedRequest.QueryResult = result
+          # }
+          # return queryCompletedRequest
+          result = executor.process_query(metadata.query_type, metadata.query_args)
+          connection.respond_query_task_completed(
+            task_token: task_token,
+            namespace: namespace,
+            completed_type: Temporal::Api::Enums::V1::QueryResultType::QUERY_RESULT_TYPE_ANSWERED,
```
antstorm, line 66: We need a way of communicating other query results, right?

antstorm (commented):
Added a few comments, but this is really great progress 🎉

examples/spec/integration/query_workflow_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'workflows/query_workflow'
+
+describe QueryWorkflow, :integration do
+  subject { described_class }
+
+  it 'returns the correct result for the queries' do
+    workflow_id, run_id = run_workflow(described_class)
```
chuckremes2, line 7: Wish I had seen this helper when I wrote tests for SignalExternal. Looks good.

examples/spec/integration/query_workflow_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'workflows/query_workflow'
+
+describe QueryWorkflow, :integration do
+  subject { described_class }
+
+  it 'returns the correct result for the queries' do
```
chuckremes2, line 6: Test currently only covers the named query "state." Can you add a test to confirm the default `on_query` handler runs? Lastly, add a test to confirm that two named `on_query` handlers can live side by side and be invoked separately?

Time permitting, I will add these tests or we can do a followup.

chuckremes2: BTW, I know the default on_query handler is implicitly tested here via the cancel behavior. But it isn't clear at first glance. I prefer explicit tests.

lib/temporal/client.rb
```diff
@@ -172,6 +172,27 @@ def signal_workflow(workflow, signal, workflow_id, run_id, input = nil, namespac
       )
     end
 
+    # Issue a query against a running workflow
+    #
+    # @param workflow [Temporal::Workflow, nil] workflow class or nil
```
chuckremes2, line 177: If this arg can be nil, would be nice to see a test that confirms the logic still works. Presumably only `workflow_id` must be populated for this to match. Can/should it raise if that arg is missing?

lib/temporal/connection/grpc.rb
```diff
@@ -420,16 +420,66 @@ def get_search_attributes
         raise NotImplementedError
       end
 
-      def respond_query_task_completed
-        raise NotImplementedError
+      def respond_query_task_completed(result:, **request_params)
+        request = Temporal::Api::WorkflowService::V1::RespondQueryTaskCompletedRequest.new(
+          **request_params.merge(query_result: to_query_payloads(result))
```
chuckremes2, line 425: Minor note but this usage of collecting keyword args doesn't match the style of the rest of this file which explicitly lists each arg. This may be a blocker if @antstorm feels strongly about keeping the style consistent if a bit more verbose.

lib/temporal/connection/grpc.rb
```diff
@@ -420,16 +420,66 @@ def get_search_attributes
...
+        client.respond_query_task_completed(request)
       end
 
       def reset_sticky_task_queue
         raise NotImplementedError
       end
 
-      def query_workflow
-        raise NotImplementedError
+      #     async _queryWorkflowHandler(input) {
```
chuckremes2, line 434: Don't forget to remove this Typescript guide...

lib/temporal/testing/local_workflow_context.rb
```diff
@@ -185,6 +185,10 @@ def on_signal(&block)
         raise NotImplementedError, 'Signals are not available when Temporal::Testing.local! is on'
       end
 
+      def on_query(&block)
```
chuckremes2, line 188: Why was this removed? There's no actual support for local query so this exception should remain.

lib/temporal/connection/grpc.rb
antstorm: Yeah, good point! The main idea was no to leak GRPC-specific stuff beyond this file. Return values do cross this boundary, but I think we should at least keep that separation for input arguments

lib/temporal/workflow/dispatcher.rb
dwillett: Can you elaborate on using the `target` to identify the specific query? I don't think I fully understand any potential implications of using a different `target`. Also if there's a design that would accommodate specific signal handler registration as well?

lib/temporal/workflow/context.rb
dwillett: Yeah I agree, hopefully we can give both signals and queries similar flexibility of handlers. Queries probably need a preference in which one is elected to be used - or warnings / errors if attempting to use both forms in the same workflow, potentially.

lib/temporal/workflow/dispatcher.rb
chuckremes2: Take a look at the context.rb for how the signals are registered and dispatched. Queries should be exactly the same. They have a target, a name, and a block which is identical to signals. So instead of tracking them in their own `query_handlers` hash, use the main one.

In context.rb, you'll register like this:
```
dispatcher.register_handler(target, query) { |input| block.call(input) }
```

To call it in executor.rb, use `dispatch` directly just like we did for signals.
```
dispatcher.dispatch(History::EventTarget.workflow, query, args)
```
The code to handle the special-case query `__cadence_web_list` would be either in task processor or executor. Don't leak it into here.

lib/temporal/workflow/executor.rb
```diff
@@ -36,6 +36,10 @@ def run
         return state_manager.commands
       end
 
+      def process_query(query, args)
```
chuckremes2, line 39: See above. Looks like dispatcher is not accessible from within task processor, so do the special query handling here.

lib/temporal/workflow/dispatcher.rb
chuckremes2: Hmmm, maybe expose a new method on this class to return the target keys.

```
def handler_keys(target)
  handlers[target].keys
end
```
This can be called in executor.rb to add `%w[__stack_trace]` to the list for the special query handle.

lib/temporal/metadata.rb
dwillett: I don't think it needs to be on metadata - just my misunderstanding of how to make that task data available. Can remove from metadata and just parse the query types/args into a struct where we need them.

lib/temporal/workflow/dispatcher.rb
chuckremes2: Hmmm, the query may conflict with event_name? I honestly don't understand this. Is there a simple example to illustrate? Otherwise, the query_handlers and handlers structures are doing the same thing as far as I can tell. You look it up by target and then match on query/event_name.

dwillett: (As far as I understand) if you were to register a `workflow.on_query('started')` it would add a handler function in `handlers[target]` with a structure of `['started', handler]`. This event name would also match and have its handler invoked for internally dispatched 'started' events that share the same target. If we want to share the data structure (with the same target lookup), something in the `event_name` would have to differentiate it as a query handler.

lib/temporal/testing/local_workflow_context.rb
dwillett: Nothing was removed here, this was added. I will need to fix the method signature on it, though

lib/temporal/workflow/dispatcher.rb
dwillett: @antstorm I'm not seeing an obvious home for reaching through to the `Dispatcher` from the `TaskProcessor` to invoke the query handler, aside from potentially delegating `dispatch` through the executor. Thoughts?

lib/temporal/workflow/dispatcher.rb
chuckremes2: Okay, I understand the issue now. Thanks for spelling it out.

My suggestion is we solve this by adopting a convention. `register_handler` is called internally where there could be conflicts in the future ('failed', 'signaled', 'completed'). Internal usages of `register_handler` should adopt a convention of two leading underscores for the event_name and then we document that for the SDK so that end users don't do the same. 

There could be a potential conflict with signal names (if we enhance `on_signal` to take a name like you did with `on_query`). It's up to the programmer to avoid name collisions but our docs could suggest a convention there where signal names start with `s_` and query names start with `q_`.

Think that gets us the result we want?

lib/temporal/connection/grpc.rb
```diff
@@ -174,12 +174,41 @@ def poll_workflow_task_queue(namespace:, task_queue:)
...
+              result_type: Temporal::Api::Enums::V1::QueryResultType::QUERY_RESULT_TYPE_FAILED,
+              error_message: query_result.message
+            }
+          else
+            {
+              result_type: Temporal::Api::Enums::V1::QueryResultType::QUERY_RESULT_TYPE_ANSWERED,
+              answer: to_query_payloads(query_result)
+            }
+          end
+        end
```
dwillett, line 205: I don't love the approach here of passing through exception objects as query results, happy to discuss alternatives

lib/temporal/workflow/dispatcher.rb
```diff
@@ -26,7 +35,7 @@ def handlers_for(target, event_name)
         handlers[target]
           .concat(handlers[TARGET_WILDCARD])
           .select { |(name, _)| name == event_name || name == WILDCARD }
-          .map(&:last)
+          .sort_by { |(name, _)| name == WILDCARD ? 1 : 0 }
```
dwillett, line 38: After looking at the existing specs, sorting wildcard handlers to the end is probably not the behavior we want for non-query use cases?

lib/temporal/workflow/dispatcher.rb
antstorm: @dwillett why not track these in a separate array in the `Workflow::Context` which you populate every time a query handler is register?

lib/temporal/workflow/context.rb
antstorm: Hmm… I have a sense that we can support both forms with the same API. I don't think there's too much room for confusion there:

```ruby
workflow.on_query('query') do |input|
```

and

```ruby
workflow.on_query do |query, input|
```

we also should be able to validate by checking block's arity when called

antstorm: Or do you think there might be confusion with `input`'s optionality?

dwillett: I sweeped through and added the catch-all query handler support in latest round of changes (within the same API). This bit of code should be outdated now.

lib/temporal/workflow/dispatcher.rb
antstorm: Can you help me understand why you need to sort handlers in the first place?

dwillett: For sure - the "catch-all" query handler registers itself with the wildcard `*` event name. This lets you use both targeted and catch-all query handlers within the same workflow (such as in the example workflow). The sorting here would reorder the wildcard catch-all handler to the end so that a specific handler takes precedence.

examples/spec/integration/query_workflow_spec.rb
dwillett: I expanded the integration test scenarios a bit in the latest round of changes

lib/temporal/workflow/context.rb
```diff
@@ -362,6 +369,11 @@ def schedule_command(command)
         state_manager.schedule(command)
       end
 
+      # @param query [Temporal::Api::Query::V1::WorkflowQuery]
+      def process_query(query)
+        dispatcher.dispatch(History::EventTarget.workflow, "queried:#{query.query_type}", query.query_args)
```
chuckremes2, line 374: This will only ever match the WILDCARD query handler. Why are you building this string with a "queried:" prefix to it?

lib/temporal/workflow/context.rb
```diff
@@ -287,6 +287,13 @@ def on_signal(&block)
         end
       end
 
+      def on_query(query = nil, &block)
+        query ||= Dispatcher::WILDCARD
```
chuckremes2, line 291: If you look at the signal implementation, it defaulted to `'signaled'` as the event_name when registering an unnamed handler. Why use the WILDCARD here?

PR #144, opened 2022-02-25
Converted WorkflowExecutionAlreadyStartedFailure handling to use RPC status

cvanderschuere:
We ran into a condition where the error message changed slightly and we were unable to extract the Run ID. This changes moves to start using the RPC error in a similar way to the [Go SDK](https://github.com/temporalio/sdk-go/blob/93e91e50709017c344a36704fd59ea51d74d26e4/internal/internal_workflow_client.go#L1054-L1061).

### DIFF
--- a/lib/temporal/connection/grpc.rb
+++ b/lib/temporal/connection/grpc.rb
@@ -5,6 +5,7 @@ require 'temporal/connection/errors'
 require 'temporal/connection/serializer'
 require 'temporal/connection/serializer/failure'
 require 'gen/temporal/api/workflowservice/v1/service_services_pb'
+require 'gen/temporal/api/errordetails/v1/message_pb'
 require 'temporal/concerns/payloads'
 
 module Temporal
@@ -117,9 +118,13 @@ module Temporal
 
         client.start_workflow_execution(request)
       rescue ::GRPC::AlreadyExists => e
-        # Feel like there should be cleaner way to do this...
-        run_id = e.details[/RunId: ([a-f0-9]+-[a-f0-9]+-[a-f0-9]+-[a-f0-9]+-[a-f0-9]+)/, 1]
-        raise Temporal::WorkflowExecutionAlreadyStartedFailure.new(e.details, run_id)
+        cast_error = e.to_rpc_status&.details&.map do |any_error|
+          next unless any_error.type_url == 'type.googleapis.com/temporal.api.errordetails.v1.WorkflowExecutionAlreadyStartedFailure'
+
+          any_error.unpack(Temporal::Api::ErrorDetails::V1::WorkflowExecutionAlreadyStartedFailure)
+        end&.compact&.first
+
+        raise Temporal::WorkflowExecutionAlreadyStartedFailure.new(e.details, cast_error&.run_id)
       end
 
       SERVER_MAX_GET_WORKFLOW_EXECUTION_HISTORY_POLL = 30
--- a/spec/unit/lib/temporal/grpc_client_spec.rb
+++ b/spec/unit/lib/temporal/grpc_client_spec.rb
@@ -4,20 +4,28 @@ describe Temporal::Connection::GRPC do
   let(:namespace) { 'test-namespace' }
   let(:workflow_id) { SecureRandom.uuid }
   let(:run_id) { SecureRandom.uuid }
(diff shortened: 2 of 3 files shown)


lib/temporal/connection/grpc.rb
```diff
@@ -117,9 +118,13 @@ def start_workflow_execution(
...
         client.start_workflow_execution(request)
       rescue ::GRPC::AlreadyExists => e
-        # Feel like there should be cleaner way to do this...
-        run_id = e.details[/RunId: ([a-f0-9]+-[a-f0-9]+-[a-f0-9]+-[a-f0-9]+-[a-f0-9]+)/, 1]
-        raise Temporal::WorkflowExecutionAlreadyStartedFailure.new(e.details, run_id)
+        cast_error = e.to_rpc_status&.details&.map do |any_error|
+          next unless any_error.type_url == 'type.googleapis.com/temporal.api.errordetails.v1.WorkflowExecutionAlreadyStartedFailure'
+
+          any_error.unpack(Temporal::Api::ErrorDetails::V1::WorkflowExecutionAlreadyStartedFailure)
+        end&.compact&.first
```
antstorm, line 125: I think this approach works, however there must be a way to unpack every error without checking it's type explicitly and then pick the right one. I have a sense that it might be useful in other places as well and this approach will not work if there's a change in the error type definition

lib/temporal/connection/grpc.rb
```diff
@@ -117,9 +118,13 @@ def start_workflow_execution(
 
         client.start_workflow_execution(request)
       rescue ::GRPC::AlreadyExists => e
-        # Feel like there should be cleaner way to do this...
-        run_id = e.details[/RunId: ([a-f0-9]+-[a-f0-9]+-[a-f0-9]+-[a-f0-9]+-[a-f0-9]+)/, 1]
-        raise Temporal::WorkflowExecutionAlreadyStartedFailure.new(e.details, run_id)
+        cast_error = e.to_rpc_status&.details&.map do |any_error|
```
antstorm, line 121: nit: `find` should work much better here and should remove the need for `map` / `compact` / `first`

spec/unit/lib/temporal/grpc_client_spec.rb
```diff
@@ -4,20 +4,28 @@
   let(:namespace) { 'test-namespace' }
   let(:workflow_id) { SecureRandom.uuid }
   let(:run_id) { SecureRandom.uuid }
-  let(:now) { Time.now}
+  let(:now) { Time.now.utc }
+  let(:already_started_error) do
```
antstorm, line 8: Will this work with the older server version? Do we have a spec to prove it?

antstorm (changes requested):
Thank you for the PR, @cvanderschuere ! Few small comments, but overall looks great 👍

lib/temporal/connection/grpc.rb
cvanderschuere: Looking at other languages, the [pattern](https://github.com/grpc/grpc-go/blob/master/examples/features/errors/client/main.go#L55-L67) seems to be to loop over and do type checks. I went ahead and moved to that form.

I also noticed there was a gap with `signal_with_start_workflow_execution` so that was updated to have this error handling

lib/temporal/connection/grpc.rb
cvanderschuere: Ended up removing `first` so that all error details are available

spec/unit/lib/temporal/grpc_client_spec.rb
cvanderschuere: What specific older version should I test against? Looking at the [proto definitions](https://github.com/temporalio/api/blame/master/temporal/api/errordetails/v1/message.proto#L45) this error has been around for at least a couple years

antstorm: @cvanderschuere I meant the one with the original unchanged message that the old code was written for

antstorm (commented):
Looks great! One last thing — can you please add a spec for raising `Temporal::ApiError`? It seems to be a new behaviour and we'd like a test for it

cvanderschuere:
@antstorm this should be ready for another review

PR #145, opened 2022-02-27
upsert_search_attributes

drewhoskins-stripe:
# Description
Adds the ability to set search attributes during a workflow's execution, to go into ElasticSearch.  
Also returns the search attributes from `fetch_workflow_execution_info`
A couple notes:
* I created Context::Private::Validators so we can easily share validations between local/production contexts
* The integration test relies on default search attributes set in the default docker compose.  circleci has them set up as well.


lib/temporal/workflow/history/event_target.rb
```diff
@@ -32,7 +33,7 @@ class UnexpectedEventType < InternalError; end
           'SIGNAL_EXTERNAL_WORKFLOW_EXECUTION'         => EXTERNAL_WORKFLOW_TYPE,
           'EXTERNAL_WORKFLOW_EXECUTION_CANCEL'         => CANCEL_EXTERNAL_WORKFLOW_REQUEST_TYPE,
           'REQUEST_CANCEL_EXTERNAL_WORKFLOW_EXECUTION' => CANCEL_EXTERNAL_WORKFLOW_REQUEST_TYPE,
-          'UPSERT_WORKFLOW_SEARCH_ATTRIBUTES'          => WORKFLOW_TYPE,
+          'UPSERT_WORKFLOW_SEARCH_ATTRIBUTES'          => UPSERT_SEARCH_ATTRIBUTES_REQUEST_TYPE,
```
drewhoskins-stripe, line 36: I confess, I don't really get what's going on with TARGET_TYPES or what I'm supposed to do.  @antstorm is this the code you thought might end up useful but hasn't yet?  
Overall, would love to see an explanatory comment framing this code if we want to keep it.

drewhoskins-stripe:
Example test failure predates this PR.

lib/temporal/workflow/history/event_target.rb
antstorm: The main idea here is to aggregate Temporal events around some target. This is then used to dispatch local events via the `Dispatcher` (where you need to use the target that is expected to receive those events).

I'll find some time to add comments to this class

spec/unit/lib/temporal/testing/local_workflow_context_spec.rb
```diff
@@ -195,5 +195,24 @@ def execute
       fiber.resume # start running
       expect(exited).to eq(true)
     end
+
+    describe '#upsert_search_attributes' do
```
antstorm, line 199: nit: missing an empty hash case

spec/unit/lib/temporal/workflow/context_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+describe Temporal::Workflow::Context do
+  let(:state_manager) { instance_double('Temporal::Workflow::StateManager') }
+  let(:dispatcher) { instance_double('Temporal::Workflow::Dispatcher') }
+  let(:metadata) { instance_double('Temporal::Metadata::Workflow') }
+  let(:workflow_context) {
+    Temporal::Workflow::Context.new(state_manager, dispatcher, MyTestWorkflow, metadata, Temporal.configuration)
+  }
+
+  describe '#upsert_search_attributes' do
```
antstorm, line 14: to make it useful we also need a spec that ensures a correct command is added when this method gets called

lib/temporal/workflow/context.rb
```diff
@@ -17,6 +17,24 @@ class Workflow
     class Context
       attr_reader :metadata, :config
 
+      module Private
+        # Shared with LocalWorkflowContext so we can do the same validations in test and production.
```
antstorm, line 21: Since it is shared we need to move it out into its own file, otherwise you're depending `LocalContext` on the full `Context`

lib/temporal/workflow/context.rb
```diff
@@ -17,6 +17,24 @@ class Workflow
...
+        class Validators
```
antstorm, line 22: nit: since it's not expected to get initialised, we should probably just use `module`

lib/temporal/testing/local_workflow_context.rb
```diff
@@ -193,6 +194,13 @@ def cancel(target, cancelation_id)
         raise NotImplementedError, 'Cancel is not available when Temporal::Testing.local! is on'
       end
 
+      def upsert_search_attributes(search_attributes)
+        Temporal::Workflow::Context::Private::Validators.validate_search_attributes(search_attributes)
+
+        # We no-op in local testing mode since there is no search functionality.  We don't fail because we 
+        # don't want to block workflows testing other aspects.
```
antstorm, line 201: nit: maybe we should still keep a track of the search attributes and allow for mocked `fetch_workflow_execution` calls to return them?

antstorm (commented):
Looks great 👍 A couple of small comments below

lib/temporal/testing/local_workflow_context.rb
drewhoskins-stripe: Thanks, mind if I punt to the next PR?  I have another fast-follow to do anyway.  Curious if you have any advice on where I should to track the state.  A class variable in `LocalWorkflowContext`, maybe?

lib/temporal/workflow/history/event_target.rb
drewhoskins-stripe: Yeah, it's a bit weird that sometimes the target is an object and sometimes a request.  Something feels not quite right.

lib/temporal/testing/local_workflow_context.rb
antstorm: Sure thing, a follow up PR sounds good.

As to where to place it — the list of executions might be more appropriate than a class variable (which would restrict parallel executions)

lib/temporal/workflow/context.rb
```diff
@@ -16,7 +17,7 @@ module Temporal
   class Workflow
     class Context
       attr_reader :metadata, :config
-
+
```
antstorm, line 20: Nit: I think your editor is not stripping trailing whitespaces, leaving these artifacts behind

lib/temporal/workflow/context_validators.rb
```diff
@@ -0,0 +1,24 @@
+
+module Temporal
+  class Workflow
+    class Context
+      module Private
+        # Shared between Context and LocalWorkflowContext so we can do the same validations in test and production.
+        module Validators
```
antstorm, line 7: Can you please match the file name/path to the module name and namespace (like we do with all the other classes/modules)?

lib/temporal/workflow/history/event_target.rb
antstorm: Yeah, it's either you split the events by target or make the state machine much more complicated. For example: cancellation request itself can fail, so if you were to track it on the activity/workflow target you'd need to somehow accommodate 2 parallel state machines

spec/unit/lib/temporal/workflow/context_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    it 'requires a non-empty hash' do
+      expect do
+        workflow_context.upsert_search_attributes({})
+      end.to raise_error(ArgumentError, 'Cannot upsert an empty hash for search_attributes, as this would do nothing.')
+    end
+
+    it 'creates a command to execute the request' do
+      allow(state_manager).to receive(:schedule)
+        .with an_instance_of(Temporal::Workflow::Command::UpsertSearchAttributes)
```
antstorm, line 35: `allow` doesn't set the expectation, just the stub. You need to add `expect().to have_received().with()` to actually ensure that it was correctly sent

drewhoskins-stripe: Whoops!

lib/temporal/workflow/context_validators.rb
drewhoskins-stripe: I guess the house style is not to use `Private` module.

lib/temporal/workflow/context.rb
drewhoskins-stripe: Ah, thanks, been meaning to fix that in this workspace.

antstorm (approved):
Awesome, thank you, @drewhoskins-stripe 🙌

PR #147, opened 2022-03-01
Improvements to upsert_search_attributes

drewhoskins-stripe:
1) Make it work in local test mode by caching the attributes on `WorkflowExecution`
2) Figured out how to pass a Datetime into temporal and updated the test and docs accordingly
3) Allow people to pass a `Time` in which gets converted under the hood to the appropriate format.

# Test Plan

Local testing: `rspec ./spec/unit/lib/temporal/testing/temporal_override_spec.rb`
Integration with datetime: 
```
# Terminal 1
cd examples
./bin/worker
# Terminal 2
rspec ./spec/integration/upsert_search_attributes_spec.rb`
```

### DIFF
--- a/examples/spec/integration/upsert_search_attributes_spec.rb
+++ b/examples/spec/integration/upsert_search_attributes_spec.rb
@@ -1,7 +1,7 @@
 require 'workflows/upsert_search_attributes_workflow'
 
-describe 'Temporal::Workflow::Context.upsert_search_attributes', :integration do 
-  it 'can upsert a search attribute and then retrieve it' do 
+describe 'Temporal::Workflow::Context.upsert_search_attributes', :integration do
+  it 'can upsert a search attribute and then retrieve it' do
     workflow_id = 'upsert_search_attributes_test_wf-' + SecureRandom.uuid
 
     expected_attributes = {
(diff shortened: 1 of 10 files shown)


examples/workflows/upsert_search_attributes_workflow.rb
```diff
@@ -1,17 +1,18 @@
 class UpsertSearchAttributesWorkflow < Temporal::Workflow
-  def execute(string_value, bool_value, float_value, int_value)
+  # time_value example: use this format: Time.now.utc.strftime("%Y-%m-%dT%H:%M:%SZ")
```
chuckremes2, line 2: ISO8601 for the win. No other time format should even be allowed. :)

chuckremes2 (approved):
I'd approve this if I could.

examples/spec/integration/upsert_search_attributes_spec.rb
```diff
@@ -1,14 +1,15 @@
...
+describe 'Temporal::Workflow::Context.upsert_search_attributes', :integration do
+  it 'can upsert a search attribute and then retrieve it' do
     workflow_id = 'upsert_search_attributes_test_wf-' + SecureRandom.uuid
 
     expected_attributes = {
       'CustomStringField' => 'moo',
       'CustomBoolField' => true,
       'CustomDoubleField' => 3.14,
       'CustomIntField' => 0,
+      'CustomDatetimeField' => Time.now.utc.strftime("%Y-%m-%dT%H:%M:%SZ"),
```
antstorm, line 12: This is part of the Ruby's standard library:

```ruby
require 'time'
Time.now.utc.iso8601
```

antstorm (commented):
Looks good! One note — I'm thinking that we should do the conversion of DateTime to string ourselves and not expect the correct format to be passed in

PR #149, opened 2022-03-02
Updating register namespace

calum-stripe:
Currently we can't specify many settings / configs that the register_namespace API provides. This PR adds the ability to set the retention period and the namespace data when registering a namespace with Temporal. 

Also this introduces an integration test (which might be controversial) which registers namespaces and tests the errors. This does create a new namespace which can't be deleted but seems like the only way to integration test this function.

### DIFF
--- a/examples/spec/integration/register_namespace_spec.rb
+++ b/examples/spec/integration/register_namespace_spec.rb
@@ -24,7 +24,6 @@ describe 'Temporal.register_namespace' do
       break if result.next_page_token == ''
 
       next_page_token = result.next_page_token
-      end
     end
 
     expect(found_namespace).to_not eq(nil)
--- a/examples/spec/integration/register_namespace_spec.rb
+++ b/examples/spec/integration/register_namespace_spec.rb
@@ -16,7 +16,7 @@ describe 'Temporal.register_namespace' do
     next_page_token = ''
 
     while found_namespace.nil?
-      result = Temporal.list_namespaces(page_size: 5, next_page_token: next_page_token)
+      result = Temporal.list_namespaces(page_size: 100, next_page_token: next_page_token)
       result.namespaces.each do |namespace|
         if namespace.namespace_info.name == name
           found_namespace = namespace
@@ -37,10 +37,10 @@ describe 'Temporal.register_namespace' do
     expect(found_namespace.config.workflow_execution_retention_ttl.seconds).to eq(retention_period * 24 * 60 * 60)
   end
 
-#   it 'errors if attempting to register a namespace with the same name' do
-#     name = "test_namespace_#{SecureRandom.uuid}"
-#     Temporal.register_namespace(name)
+  it 'errors if attempting to register a namespace with the same name' do
+    name = "test_namespace_#{SecureRandom.uuid}"
+    Temporal.register_namespace(name)
     
-#     expect {Temporal.register_namespace(name)}.to raise_error(Temporal::NamespaceAlreadyExistsFailure, 'Namespace already exists.')
-#   end
+    expect {Temporal.register_namespace(name)}.to raise_error(Temporal::NamespaceAlreadyExistsFailure, 'Namespace already exists.')
+  end
 end
--- a/examples/spec/integration/register_namespace_spec.rb
+++ b/examples/spec/integration/register_namespace_spec.rb
@@ -6,30 +6,27 @@ describe 'Temporal.register_namespace' do
     name = "test_namespace_#{SecureRandom.uuid}"
     description = 'this is the description'
     retention_period = 30
-    namespace_data = { test: 'value' }
+    data = { test: 'value' }
 
-    result = Temporal.register_namespace(name, description, retention_period: retention_period, namespace_data: namespace_data)
-    expect(result).to be_an_instance_of(Temporal::Api::WorkflowService::V1::RegisterNamespaceResponse)
+    Temporal.register_namespace(name, description, retention_period: retention_period, data: data)
 
-    # fetch the namespace from Temporal and check it exists and has the correct settings
-    found_namespace = nil
-    next_page_token = ''
-
-    while found_namespace.nil?
-      result = Temporal.list_namespaces(page_size: 100, next_page_token: next_page_token)
-      found_namespace = result.namespaces.find do |namespace|
-        namespace.namespace_info.name == name
+    # fetch the namespace from Temporal and check it exists and has the correct settings 
+    # (need to wait a few seconds for temporal to catch up so try a few times)
+    attempts = 0
+    while attempts < 25 do
+      attempts += 1
+      
+      begin
+        result = Temporal.describe_namespace(name)
+         
+        expect(result.namespace_info.name).to eq(name)
+        expect(result.namespace_info.data).to eq(data)
+        expect(result.config.workflow_execution_retention_ttl.seconds).to eq(retention_period * 24 * 60 * 60)
+        break
+      rescue GRPC::NotFound
+        sleep 2
       end
-
-      break if result.next_page_token == ''
-
-      next_page_token = result.next_page_token
     end
-
-    expect(found_namespace).to_not eq(nil)
-    expect(found_namespace.namespace_info.name).to eq(name)
-    expect(found_namespace.namespace_info.data).to eq(namespace_data)
-    expect(found_namespace.config.workflow_execution_retention_ttl.seconds).to eq(retention_period * 24 * 60 * 60)
   end
 
   it 'errors if attempting to register a namespace with the same name' do
--- a/examples/spec/integration/register_namespace_spec.rb
+++ b/examples/spec/integration/register_namespace_spec.rb
@@ -1,0 +1,46 @@
+require 'workflows/timeout_workflow'
+
+describe 'Temporal.register_namespace' do
+  it 'can register a new namespace' do
+    # have to generate a new namespace on each run because currently can't delete namespaces
+    name = "test_namespace_#{SecureRandom.uuid}"
+    description = 'this is the description'
+    retention_period = 30
+    namespace_data = { test: 'value' }
+
+    result = Temporal.register_namespace(name, description, retention_period: retention_period, namespace_data: namespace_data)
+    expect(result).to be_an_instance_of(Temporal::Api::WorkflowService::V1::RegisterNamespaceResponse)
+
+    # fetch the namespace from Temporal and check it exists and has the correct settings
+    found_namespace = nil
+    next_page_token = ''
+
+    while found_namespace.nil?
+      result = Temporal.list_namespaces(page_size: 5, next_page_token: next_page_token)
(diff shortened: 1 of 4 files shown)


examples/spec/integration/register_namespace_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+    found_namespace = nil
+    next_page_token = ''
+
+    while found_namespace.nil?
+      result = Temporal.list_namespaces(page_size: 100, next_page_token: next_page_token)
+      result.namespaces.each do |namespace|
+        if namespace.namespace_info.name == name
+          found_namespace = namespace
+          break
+        end
```
chuckremes2, line 24: if `namespaces` is enumerable, then just use `find` directly, e.g.
```suggestion
      found_namespace = result.namespaces.find do |namespace|
        namespace.namespace_info.name == name
        end
```

examples/spec/integration/register_namespace_spec.rb
```diff
@@ -0,0 +1,46 @@
+require 'workflows/timeout_workflow'
+
+describe 'Temporal.register_namespace' do
+  it 'can register a new namespace' do
```
chuckremes2, line 4: This tests 4 separate things. Why not break these into individual tests so we can confirm a) it's found, b) the name is correct, c) the data matches, and d) the retention seconds are correct?

examples/spec/integration/register_namespace_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        break
+      else
+        next_page_token = result.next_page_token
+      end
+    end
+
+    expect(found_namespace).to_not eq(nil)
+    expect(found_namespace.namespace_info.name).to eq(name)
+    expect(found_namespace.namespace_info.data).to eq(namespace_data)
+    expect(found_namespace.config.workflow_execution_retention_ttl.seconds).to eq(retention_period * 24 * 60 * 60)
```
chuckremes2, line 37: If behind the scenes the retention TTL is converted to seconds, why not try and pass a small float value so that retention seconds is under 5; then you could test to see if the namespace disappears after the retention seconds expire.

I know the method signature says it takes "integer days" as the arg, but there's no reason you can't break that for testing purposes. Plus, there's no code here to a) confirm the arg is an integer, or b) convert it to an integer. That said, 0 is also an integer so that would make the namespace expire immediately, yes?

lib/temporal/connection/grpc.rb
```diff
@@ -31,14 +31,15 @@ def initialize(host, port, identity)
         @poll_request = nil
       end
 
-      def register_namespace(name:, description: nil, global: false, retention_period: 10)
+      def register_namespace(name:, description: nil, global: false, retention_period: 10, namespace_data: nil)
```
chuckremes2, line 34: Code seems fine, but I don't see why `retention_period` has to be "days".

lib/temporal/connection/grpc.rb
```diff
@@ -31,14 +31,15 @@ def initialize(host, port, identity)
...
         request = Temporal::Api::WorkflowService::V1::RegisterNamespaceRequest.new(
           namespace: name,
           description: description,
           is_global_namespace: global,
           workflow_execution_retention_period: Google::Protobuf::Duration.new(
             seconds: retention_period * 24 * 60 * 60
```
chuckremes2, line 40: This seems like a latent bug. Should wrap this whole thing and call `to_i` on it to convert to integer.

examples/spec/integration/register_namespace_spec.rb
calum-stripe: nifty, sorry ruby isn't my main language! love this tho and will commit it

lib/temporal/connection/grpc.rb
calum-stripe: I guess legacy reasons, happy to update this but don't want to break any existing functionality if we do

lib/temporal/connection/grpc.rb
calum-stripe: Sorry wrap this whole thing in what?

examples/spec/integration/register_namespace_spec.rb
calum-stripe: Yeah it's not the namespace TTL but rather how long Temporal will keep workflow histories after they've finished running. I could add this test but it feels like it's testing an implementation detail of Temporal (we're testing an internal mechanic of the server). I know the case could be made that is what integration tests are but I invisigned more along the lines:
1. Check that the server can actually accept the request
2. Check the server responds with sensible values
Rather than testing exactly what the server implementation does with the config! Open to discussion here tho

lib/temporal/connection/grpc.rb
chuckremes2: Based on your other response that this is how long workflow events are retained, then I withdraw my comment. I thought it was namespace expiration.

lib/temporal/connection/grpc.rb
chuckremes2: Heh, it's a Ruby thing. If you do `(3.14159).to_i` it returns `3`. It just converts floats to ints, that's all.
```suggestion
            seconds: (retention_period * 24 * 60 * 60).to_i
```

examples/spec/integration/register_namespace_spec.rb
chuckremes2: Ha, just learned something. Okay, I guess this part doesn't matter.

lib/temporal/connection/grpc.rb
calum-stripe: ah thank you!

drewhoskins-stripe (changes requested):
based on comments in the forked branch.

lib/temporal/client.rb
```diff
@@ -132,14 +132,13 @@ def schedule_workflow(workflow, cron_schedule, *input, options: {}, **args)
...
-    # @param global [Boolean] flag to set if the namespace is a global namespace or not
-    # @param retention_period [Int] optional retention period in days
-    # @param namespace_data [Hash] A key-value map for any customized purpose
-    def register_namespace(name, description = nil, global: false, retention_period:  10, namespace_data: nil)
-      connection.register_namespace(name: name, description: description, global: global, retention_period: retention_period, namespace_data: namespace_data)
+    # @param is_global [Boolean] used to distinguish local namespaces from global namespaces (https://docs.temporal.io/docs/server/namespaces/#global-namespaces)
+    # @param retention_period_days [Int] optional  value which specifies how long Temporal will keep workflows after completing
+    # @param namespace_data [Hash] optional key-value map for any customized purpose that can be retreived with describe_namespace
+    def register_namespace(name, description = nil, is_global: false, retention_period_days:  10, data: nil)
+      connection.register_namespace(name: name, description: description, is_global: is_global, retention_period_days: retention_period_days, data: data)
```
chuckremes2, line 139: By changing `global` to `is_global` this is a breaking change. Is the rename really worth it? I'd say no.

Same for the other arguments `retention_period` and `namespace_data`. 

I think this is best handled by improving the documentation on the header as you have done. The argument changes will cause breakage on any upgrade and I'm not convinced the new names are that much better.

@antstorm opinion?

calum-stripe: Hey, no API changed here (so no breaking change). I think the confusion is previously in this PR this param was called `global` but before this PR, `global` never existed here! Happy to chat if it's something else?

lib/temporal/client.rb
```diff
@@ -132,8 +132,11 @@ def schedule_workflow(workflow, cron_schedule, *input, options: {}, **args)
     #
     # @param name [String] name of the new namespace
     # @param description [String] optional namespace description
-    def register_namespace(name, description = nil)
-      connection.register_namespace(name: name, description: description)
+    # @param is_global [Boolean] used to distinguish local namespaces from global namespaces (https://docs.temporal.io/docs/server/namespaces/#global-namespaces)
+    # @param retention_period_days [Int] optional  value which specifies how long Temporal will keep workflows after completing
+    # @param namespace_data [Hash] optional key-value map for any customized purpose that can be retreived with describe_namespace
```
antstorm, line 137: nit: typo, should be `data` based on the method signsture

examples/spec/integration/register_namespace_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+describe 'Temporal.register_namespace' do
+  it 'can register a new namespace' do
+    # have to generate a new namespace on each run because currently can't delete namespaces
+    name = "test_namespace_#{SecureRandom.uuid}"
+    description = 'this is the description'
+    retention_period_days = 30
+    data = { test: 'value' }
+
+    Temporal.register_namespace(name, description, retention_period_days: retention_period_days, data: data)
```
antstorm, line 11: We probably need to cleanup this as it will pollute people's local Temporal setup with every test run

examples/spec/integration/register_namespace_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'workflows/timeout_workflow'
```
antstorm, line 1: Why do we need this here?

examples/spec/integration/register_namespace_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    description = 'this is the description'
+    retention_period_days = 30
+    data = { test: 'value' }
+
+    Temporal.register_namespace(name, description, retention_period_days: retention_period_days, data: data)
+
+    # fetch the namespace from Temporal and check it exists and has the correct settings 
+    # (need to wait a few seconds for temporal to catch up so try a few times)
+    attempts = 0
+    while attempts < 30 do
```
antstorm, line 16: Are we sure 15s is enough? Also, how long does it normally take?

lib/temporal/client.rb
antstorm: Yes, there should be no breaking changes since GRPC interface is private. As for adding `_days` I think this needs to be addressed across more broadly (if this is the pattern we'd like to follow) and not individually

antstorm (changes requested):

examples/spec/integration/register_namespace_spec.rb
calum-stripe: normally takes 3-5seconds and i ran this like 100 times and it didn't fail!

examples/spec/integration/register_namespace_spec.rb
calum-stripe: You can't [delete namespaces ](https://community.temporal.io/t/how-to-delete-a-namespace/677)(at least not currently supported). Any ideas how we can clean it up?

antstorm (approved):
Looks great! 👍

examples/spec/integration/register_namespace_spec.rb
antstorm: Oh wow, that's unfortunate. Let's make sure to follow up once Temporal has implemented it

PR #152, opened 2022-03-03
[Fix] Remove deprecation warning

### LINKED ISSUE #143: Deprecation message spam
Hey folks, I'm just dipping my toes into temporal and this SDK is a lifesaver, thanks for open sourcing it!

I'm playing with temporal locally and I'm seeing a lot of deprecation message spam when running workflows/activities:

```
temporal-poc-app-1      | [DEPRECATION] This method is now deprecated without a substitution
temporal-poc-app-1      | [DEPRECATION] This method is now deprecated without a substitution
temporal-poc-app-1      | [DEPRECATION] This method is now deprecated without a substitution
temporal-poc-app-1      | [DEPRECATION] This method is now deprecated without a substitution
temporal-poc-app-1      | [DEPRECATION] This method is now deprecated without a substitution
```

This appears to be due to [this line](https://github.com/coinbase/temporal-ruby/blob/e249d63cdbdd898ff410c267e8ce51286db982e4/lib/temporal.rb#L36).

Which is still used in multiple spots in the codebase:
- [here](https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/concerns/payloads.rb#L55)
- [here](https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/client.rb#L283)
- [and here](https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/worker.rb#L12)

Do you have any guidance for removing the deprecated code?  I'm happy to attempt a PR, but wanted to check with the maintainers first.

antstorm:
Fixes #143 

After #92 was merged we've added a deprecation warning to restrict references to `Temporal.configuration` from the code. However we still rely on it heavily from `Temporal::Concerns::Payloads` that is included in quite a few places. This PR removes this issue in favour of explicit dependency injection.

The meat of this PR is the introduction of `Temporal::ConverterWrapper` (that now houses all the convenience methods from `Temporal::Concerns::Payloads`) and passing it explicitly where conversion functionality is needed.

Ideally in the future we should limit conversion to the our API wrapper (`Connection::GRPC`) and not allow any proto objects beyond that layer. But this is a much bigger refactor (many proto messages would require a local class).

Tested using existing unit & integration specs. New unit specs added to cover added functionality.


spec/fabricators/grpc/history_event_fabricator.rb
```diff
@@ -130,7 +125,7 @@ class TestSerializer
   event_type { Temporal::Api::Enums::V1::EventType::EVENT_TYPE_ACTIVITY_TASK_CANCELED }
   activity_task_canceled_event_attributes do |attrs|
     Temporal::Api::History::V1::ActivityTaskCanceledEventAttributes.new(
-      details: TestSerializer.to_details_payloads('ACTIVITY_ID_NOT_STARTED'),
+      details: Temporal::Configuration.new.converter.to_details_payloads('ACTIVITY_ID_NOT_STARTED'),
```
chuckremes2, line 128: Shouldn't this be `$converter.to_details_payloads(...)`? Not sure why you are allocating a new one here.

chuckremes2 (commented):
Looks pretty good as a draft. Most of the changes were pretty repetitive so no comments on those. Main focus is probably on spec/unit/lib/temporal/converter_wrapper_spec.rb since that's where the ConverterWrapper gets tested. Looks complete to me.

I think the move to dependency injection is a good idea. I like the concept overall.

spec/fabricators/grpc/history_event_fabricator.rb
antstorm: You're right, forgot to replace this one 👍

0xTheProDev:
Is there any update here? @antstorm @chuckremes2

santiagodoldan:
It would be awesome to merge this one 💯

arrowcircle:
Any updates on this?

cdimitroulas:
What is needed to get this over the line? It would be great to finally have a fix which removes the spam of log messages from `temporal-ruby` in our applications 🙏

ukd1:
Ping! Do y'all want help on this?

antstorm:
@cdimitroulas @ukd1 @arrowcircle @santiagodoldan @0xTheProDev I've finally found some time and  re-implemented this PR here — #314

PR #154, opened 2022-03-04
Fix bugs with workflow.upsert_search_attributes

drewhoskins-stripe:
# Description
1) Fixes a bug in lib/temporal/workflow/history/event.rb where we tried to access `scheduled_event_id` on `UpsertWorkflowSearchAttributesEventAttributes`, which doesn't exist.
2) Fix a replay bug in `lib/temporal/workflow/state_manager.rb` where we weren't moving past the `UpsertWorkflowSearchAttributes` command in the workflow history.

# Test Plan
Add a subsequent activity in `UpsertSearchAttributesWorkflow` to make sure the replay code kicks in, increasing test coverage, then run

```
# Terminal 1
./bin/workers
# Terminal 2
rspec ./spec/integration/upsert_search_attributes_spec.rb
```


antstorm (approved):
Looks good, totally missed this when reviewing the original PR. I'll add comments to Event and EventTarget classes to demystify some of these concepts

PR #157, opened 2022-03-07
Named signal handlers

chuckremes2:
Support for named signal handlers, e.g. `on_signal('SOME_NAME') { }`. Always call the "catch-all" signal handler even when a named handler exists.

### DIFF
--- a/lib/temporal/workflow/context.rb
+++ b/lib/temporal/workflow/context.rb
@@ -280,11 +280,23 @@ module Temporal
         state_manager.local_time
       end
 
-      def on_signal(&block)
+      # Define a signal handler to receive signals onto the workflow. When
+      # +name+ is defined, this creates a named signal handler which will be
+      # invoked whenever a signal named +name+ is received. A handler without
+      # a set name (defaults to nil) will be the default handler and will receive
+      # all signals that do not match a named signal handler.
+      #
+      # @param name [String, Symbol, nil] an optional signal name; converted to a String
+      def on_signal(handler_name=nil, &block)
         target = History::EventTarget.workflow
 
-        dispatcher.register_handler(target, 'signaled') do |signal, input|
-          call_in_fiber(block, signal, input)
+        dispatcher.register_handler(target, 'signaled', handler_name: handler_name) do |signal, input|
+          if handler_name
+            # do not pass signal name when triggering a named handler
+            call_in_fiber(block, input)
+          else
+            call_in_fiber(block, signal, input)
+          end
         end
       end
 
--- a/lib/temporal/workflow/context.rb
+++ b/lib/temporal/workflow/context.rb
@@ -297,7 +297,7 @@ module Temporal
             call_in_fiber(block, input)
           end
         else
-          dispatcher.register_handler(target, Dispatcher::WILDCARD) do |signal, input|
+          dispatcher.register_handler(Dispatcher::TARGET_WILDCARD, 'signaled') do |signal, input|
             call_in_fiber(block, signal, input)
           end
         end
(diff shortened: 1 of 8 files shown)


lib/temporal/workflow/dispatcher.rb
```diff
@@ -1,19 +1,38 @@
...
+    # points to this class, #register_handler and #dispatch.
+    #
+    # A handler may be associated with a specific event name so when that event occurs
+    # elsewhere in the system we may dispatch the event and execute the handler. By default
+    # we do a simple name <=> handler association.
+    #
+    # Optionally, we may register a named handler that is triggered when an event _and
+    # an optional handler_name key_ are provided. When this more specific match
+    # fails, we fall back to matching just against +event_name+.
+    #
```
antstorm, line 13: I feel like a cleaner solution would be to add the `handler_name` to `event_name` and dispatch the event twice:
- "signaled:signalName"
- "signaled"

I think this would keep the dispatcher simpler for now and does allow us to achieve everything that is needed

antstorm (changes requested):
See inline comments

lib/temporal/workflow/dispatcher.rb
```diff
@@ -22,11 +37,24 @@ def dispatch(target, event_name, args = nil)
...
+          .map(&:handler)
+      end
+
+      def match?(event_struct, event_name, named_handler)
+        event_struct.event_name == event_name ||
+          event_struct.event_name == named_handler ||
+          event_struct.event_name == WILDCARD
+      end
+
+      def sanitize_handler_name(event_name, handler_name)
```
antstorm, line 55: I wonder if we should let the called to this instead of handling it in the Dispatcher ^. Basically I'm thinking that we can leave the Dispatcher's API unchanged and just allow the caller to construct the event name the way they want.

Another approach might be to actually use the target to represent a signal itself, then what you can do is:

```ruby
signal = Signal.new(signal_name)
dispatcher.register_handler(signal, 'signaled') { |input| ... }
```

and for a catch-all:

```ruby
dispatcher.register_handler(Dispatcher::TARGET_WILDCARD, 'signaled') { |signal, input| ... }
```

I'm a bit hesitant to extend the Dispatcher's interface because I feel like we can achieve what we need with the current version and can avoid overcomplicating it

chuckremes2: As you see, I tried to move the naming logic into Dispatcher. I agree that it changes the Dispatcher interface but now the complexity will move elsewhere.

I like the second suggestion about making a `Signal` object. I'll try that first since it preserves some encapsulation without leaking it to multiple classes.

lib/temporal/workflow/context.rb
```diff
@@ -280,11 +280,26 @@ def now
...
-        dispatcher.register_handler(target, 'signaled') do |signal, input|
-          call_in_fiber(block, signal, input)
+        if signal_name
+          event_name = "signaled:#{signal_name}"
+          dispatcher.register_handler(target, event_name) do |_, input|
+            # do not pass signal name when triggering a named handler
+            call_in_fiber(block, input)
+          end
+        else
+          dispatcher.register_handler(target, Dispatcher::WILDCARD) do |signal, input|
```
antstorm, line 300: Something is missing here. This will basically call the handler on every single event dispatched on the `History::EventTarget.workflow`

chuckremes2: I was worried about that but unsure. You confirmed it was wrong. Just pushed a change more in line with your earlier guidance using `Dispatcher::TARGET_WILDCARD, 'signaled'` instead. Hopefully that won't get executed for all workflows whenever a signal arrives.

PR #161, opened 2022-03-08
Added fix for nil search attributes when listing workflows

calum-stripe:
list_workflow_executions is currently not working when search attributes object is nil. This PR adds the nil check for `response.search_attributes` to ensure it doesn't blow up.

### DIFF
--- a/spec/unit/lib/temporal/workflow/execution_info_spec.rb
+++ b/spec/unit/lib/temporal/workflow/execution_info_spec.rb
@@ -25,7 +25,7 @@ describe Temporal::Workflow::ExecutionInfo do
       api_info.search_attributes = nil
 
       result = described_class.generate_from(api_info)
-      expect(result).to be_an_instance_of(Temporal::Workflow::ExecutionInfo)
+      expect(result.search_attributes).to be_nil
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/unit/lib/temporal/workflow/execution_info_spec.rb
```diff
@@ -20,6 +20,13 @@
     it 'freezes the info' do
       expect(subject).to be_frozen
     end
+
+    it 'deserializes if search_attributes is nil' do
+      api_info.search_attributes = nil
+
+      result = described_class.generate_from(api_info)
+      expect(result.search_attributes).to be_nil
```
antstorm, line 28: Feels like this should be an empty hash and not `nil` (assuming there's no semantic difference). This would prevent SDK users from having to do their own nil-checks further down

antstorm (changes requested):

spec/unit/lib/temporal/workflow/execution_info_spec.rb
calum-stripe: Fixed, let me know what you think!

DeRauk: @antstorm I'm going to approve and merge since this comment seems to be resolved, but let me know if there were any other changes you wanted and I can open a fix PR.

DeRauk: @antstorm I'm going to approve and merge since this comment seems to be resolved, but let me know if there were any other changes you wanted and I can open a fix PR.

DeRauk (approved):

PR #164, opened 2022-03-11
Expose scheduled_time and current_attempt_scheduled_time on activity metadata

nagl-stripe:
## Summary
This adds two scheduling-related fields to the activity metadata. These are present in the temporal java SDK.

## Testing
```
bundle exec rspec spec/unit/

cd examples
./bin/worker &
USE_ENCRYPTION=1 ./bin/worker &
```

### DIFF
--- a/lib/temporal/metadata.rb
+++ b/lib/temporal/metadata.rb
@@ -22,9 +22,8 @@ module Temporal
           workflow_name: task.workflow_type.name,
           headers: from_payload_map(task.header&.fields || {}),
           heartbeat_details: from_details_payloads(task.heartbeat_details),
-          # temporal doesn't render sub-second times, so we ignore the nanos field
           scheduled_at: task.scheduled_time.to_time,
-          current_attempt_scheduled_at: task.current_attempt_scheduled_time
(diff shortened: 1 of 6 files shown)


antstorm (changes requested):
Thanks @nagl-stripe 👍 

Couple of improvements for this to get merged:

1. Let's return the `DateTime` objects for time values like we do with the other metadata (e.g. `lib/temporal/metadata.rb:49`)
2. Following Rails' convention, let's also use `_at` in the field's name to signify a `DateTime` result

Hope that makes sense!

lib/temporal/metadata.rb
```diff
@@ -21,7 +21,10 @@ def generate_activity_metadata(task, namespace)
           workflow_id: task.workflow_execution.workflow_id,
           workflow_name: task.workflow_type.name,
           headers: from_payload_map(task.header&.fields || {}),
-          heartbeat_details: from_details_payloads(task.heartbeat_details)
+          heartbeat_details: from_details_payloads(task.heartbeat_details),
+          # temporal doesn't render sub-second times, so we ignore the nanos field
```
antstorm, line 25: This comment is probably now irrelevant

antstorm (approved):
🙌

PR #167, opened 2022-03-18
Expose wait_for_start for child workflow execution

arya-k:
# Summary
Currently, in order to ensure child workflows behave as expected, we have to wait for temporal to send a ChildExecutionStarted history event. Since we can't do that, we've been sleeping. This PR adds a wait_for_start to child workflow execution so that we can reliably wait for Child workflows, as [indicated in Temporal's docs](https://docs.temporal.io/docs/concepts/what-is-a-child-workflow-execution)

# Testing
I ran:
```
bundle exec rspec spec/
cd examples
./bin/worker
USE_ENCRYPTION=1 ./bin/worker
```

I modified parent_close_workflow_spec to use this flag instead of sleep. If you disable `wait_for_start`, the test will fail nondeterministically.

### DIFF
--- a/lib/temporal/workflow/context.rb
+++ b/lib/temporal/workflow/context.rb
@@ -103,6 +103,7 @@ module Temporal
         input << args unless args.empty?
 
         parent_close_policy = options.delete(:parent_close_policy)
+        wait_for_start = options.delete(:wait_for_start)
         execution_options = ExecutionOptions.new(workflow_class, options, config.default_execution_options)
 
         command = Command::StartChildWorkflow.new(
@@ -129,6 +130,14 @@ module Temporal
         dispatcher.register_handler(target, 'failed') do |exception|
           future.fail(exception)
           future.failure_callbacks.each { |callback| call_in_fiber(callback, exception) }
+        end
+
+        if wait_for_start
+          child_workflow_started = false
+          dispatcher.register_handler(target, 'started') do
+            child_workflow_started = true
+          end
+          wait_for { child_workflow_started }
         end
 
         future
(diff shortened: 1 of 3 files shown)


lib/temporal/workflow/state_manager.rb
```diff
@@ -241,6 +241,7 @@ def apply_event(event)
           dispatch(target, 'failed', 'StandardError', from_payloads(event.attributes.cause))
 
         when 'CHILD_WORKFLOW_EXECUTION_STARTED'
+          dispatch(target, 'started', event.attributes.workflow_execution)
```
antstorm, line 244: There seems to be an issue here — you're dispatching the event with one argument (`event.attributes.workflow_execution`), however the handler is registered without arguments. This suggests that the new behaviour isn't working properly and we need to figure out why the tests are still passing (indicating a potential flake)

antstorm (changes requested):
@arya-k thanks for fixing this, really appreciate it 👍 Left 2 comments, please let me know if that doesn't make sense for any reason

lib/temporal/workflow/context.rb
```diff
@@ -103,6 +103,7 @@ def execute_workflow(workflow_class, *input, **args)
         input << args unless args.empty?
 
         parent_close_policy = options.delete(:parent_close_policy)
+        wait_for_start = options.delete(:wait_for_start)
```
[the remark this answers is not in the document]
arya-k: I agree– I'll promote this behavior to always be enabled, with a comment indicating that temporal says we *must* wait for the child to get spawned.

lib/temporal/workflow/state_manager.rb
antstorm: @arya-k hmm, interesting:

```ruby
>> proc = -> { p 'called' }
=> #<Proc:0x0000000123aa6da0 (irb):1 (lambda)>
>> proc.call('a')
(irb):1:in `block in <main>': wrong number of arguments (given 1, expected 0) (ArgumentError)
```

I wonder if a block is handled differently. But you're right, your example works as expected and just ignores the argument

PR #172, opened 2022-04-06
Fix bugs around child workflow ids and workflow_reuse_policy

nagl-stripe:
### Summary
This PR fixes two bugs around workflow id reuse with child workflows:
* Pass workflow_id_reuse_policy through (instead of dropping it)
* Render errors correctly when the child workflow can't be started (the old code treats them like payloads, which they aren't)

Read commit-by-commit for a logical breakdown of the changes.

### Testing
There's one new suite of integration tests: spec/integration/child_id_reuse_spec.rb.

I ran:
```
# Units...
bundle exec rspec spec/unit

# Integs...
cd examples
./bin/worker &
USE_ENCRYPTION=1 ./bin/worker &
```


spec/unit/lib/temporal/grpc_client_spec.rb
```diff
@@ -28,12 +28,61 @@
...
-          memo: {}
+          memo: {},
+          workflow_id_reuse_policy: :allow,
         )
       end.to raise_error(Temporal::WorkflowExecutionAlreadyStartedFailure) do |e|
         expect(e.run_id).to eql('baaf1d86-4459-4ecd-a288-47aeae55245d')
       end
     end
+
+    it 'starts a workflow with scalar arguments' do
```
nagl-stripe, line 39: There was a test like this one for signal_with_start, but there wasn't one for start. I wanted to add workflow_id_reuse_policy to the args/asserts for both call paths.

lib/temporal/errors.rb
```diff
@@ -57,10 +57,11 @@ class WorkflowAlreadyCompletingError < InternalError; end
   class WorkflowExecutionAlreadyStartedFailure < ApiError
     attr_reader :run_id
 
-    def initialize(message, run_id)
+    def initialize(message, run_id = nil)
```
nagl-stripe, line 60: This is more interesting than it looks: in order for the error to deserialize correctly from await_workflow_result, it needs to have a 1-arg constructor. We don't have the run_id available when a child workflow fails to start.

Defaulting to nil seemed like a good way to solve both of those problems at once.

drewhoskins-stripe: Odd that they don't send back the run_id for child workflows.

lib/temporal/workflow/errors.rb
```diff
@@ -52,6 +52,17 @@ def self.generate_error(failure, default_exception_class = StandardError)
         end
       end
 
+      def self.generate_error_for_child_workflow_start(event)
+        if event.attributes.cause == :START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_WORKFLOW_ALREADY_EXISTS
+          Temporal::WorkflowExecutionAlreadyStartedFailure.new(
+            "The child workflow could not be started because a workflow with its id already exists: #{event.attributes.workflow_id}",
```
drewhoskins-stripe, line 58: You might want to mention the mutual exclusion policy.

drewhoskins-stripe (commented):
Nice, thanks for polishing child workflows!

lib/temporal/workflow/errors.rb
nagl-stripe: Like the run_id, it isn't available in this context / stack frame. I could use the start_child_workflow history event - it's available there. Do you think it's worth the factoring necessary to do so?

lib/temporal/connection/serializer/workflow_id_reuse_policy.rb
```diff
@@ -0,0 +1,25 @@
...
+    module Serializer
+      class WorkflowIdReusePolicy
+
+        WORKFLOW_ID_REUSE_POLICY = {
+          allow_failed: Temporal::Api::Enums::V1::WorkflowIdReusePolicy::WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE_FAILED_ONLY,
+          allow: Temporal::Api::Enums::V1::WorkflowIdReusePolicy::WORKFLOW_ID_REUSE_POLICY_ALLOW_DUPLICATE,
+          reject: Temporal::Api::Enums::V1::WorkflowIdReusePolicy::WORKFLOW_ID_REUSE_POLICY_REJECT_DUPLICATE
+        }.freeze
+
+        def self.to_proto(reuse_policy_sym)
```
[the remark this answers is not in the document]
nagl-stripe: I'll give that a try. This is the part I was the least sure of - ty for the suggestion.

examples/spec/integration/child_id_reuse_spec.rb
```diff
@@ -0,0 +1,105 @@
+require 'workflows/parent_id_reuse_workflow'
```
antstorm, line 1: nit: can you please match the name of the file to the name of the workflow that is being tested?

lib/temporal/workflow/context.rb
```diff
@@ -116,6 +116,7 @@ def execute_workflow(workflow_class, *input, **args)
           timeouts: execution_options.timeouts,
           headers: execution_options.headers,
           memo: execution_options.memo,
+          workflow_id_reuse_policy: options[:workflow_id_reuse_policy]
```
antstorm, line 119: nit: can you please make it similar to `parent_close_policy`? This is a very similar behaviour, no reason for different implementation

lib/temporal/workflow/context.rb
```diff
@@ -136,7 +137,8 @@ def execute_workflow(workflow_class, *input, **args)
         dispatcher.register_handler(target, 'started') do
           child_workflow_started = true
         end
-        wait_for { child_workflow_started }
+
+        wait_for { child_workflow_started || future.failed? }
```
antstorm, line 141: nice catch!

lib/temporal/workflow/errors.rb
```diff
@@ -52,6 +52,17 @@ def self.generate_error(failure, default_exception_class = StandardError)
         end
       end
 
+      def self.generate_error_for_child_workflow_start(event)
+        if event.attributes.cause == :START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_WORKFLOW_ALREADY_EXISTS
```
antstorm, line 56: `START_CHILD_WORKFLOW_EXECUTION_FAILED_CAUSE_WORKFLOW_ALREADY_EXISTS` is probably a constant in the protobuf, can you please reference it instead of a hardcoded symbol?

antstorm (changes requested):
Suggested a couple of improvements, otherwise looks good 👍 Thank you, @nagl-stripe

lib/temporal/workflow/errors.rb
nagl-stripe: It's mildly cumbersome - the protobuf constant is an Integer, and you have to call an extra method to resolve it to a symbol. I'll post the change - lmk if you want me to put it back the old way.

antstorm (approved):
Looks good, thank you @nagl-stripe !

PR #181, opened 2022-04-21
Child workflow futures

calum-stripe:
# Summary
Adds a future to the `execute_workflow` future, which will keep track of whether or not the child workflow has been started yet. This is very similar to the Goland SDK which uses [GetChildWorkflowExecution](https://docs.temporal.io/docs/go/how-to-spawn-a-child-workflow-execution-in-go/) 

This PR is an improvement of https://github.com/coinbase/temporal-ruby/pull/170

# Test Plan
```
# unit test
bundle exec rspec spec/

# integration test
cd examples
(terminal 1) ./bin/worker
```


lib/temporal/workflow/context.rb
```diff
@@ -133,15 +137,16 @@ def execute_workflow(workflow_class, *input, **args)
...
           future.fail(exception)
           future.failure_callbacks.each { |callback| call_in_fiber(callback, exception) }
-        end
 
-        # Temporal docs say that we *must* wait for the child to get spawned:
-        child_workflow_started = false
-        dispatcher.register_handler(target, 'started') do
-          child_workflow_started = true
+          # child_workflow_execution_future.fail(exception)
+          # child_workflow_execution_future.failure_callbacks.each { |callback| call_in_fiber(callback, exception) }
```
antstorm, line 142: Is this needed?

lib/temporal/workflow/future.rb
```diff
@@ -4,6 +4,7 @@ module Temporal
   class Workflow
     class Future
       attr_reader :target, :success_callbacks, :failure_callbacks
+      attr_accessor :child_workflow_execution_future # will complete when the child workflow has been completed
```
antstorm, line 7: This is specific to one case, but extends the interface for all Futures. Can we crease a subclass for child workflow future that gets initialized with this value? (So we don't need a writer exposed)

antstorm (changes requested):
Looks good! Left couple of inline comments.

I also wonder what happens if the workflow fails to start? Is that even a case? Can a child workflow execution future fail?

antstorm (commented):
There's another thing that we need to bring over from Go SDK — the child workflow execution future needs to return the workflow execution object for the child workflow. I'm not sure what is it initialized with right now and we need a unit test to confirm

lib/temporal/workflow/future.rb
calum-stripe: Agreed, added this!

calum-stripe:
>the child workflow execution future needs to return the workflow execution
added and a unit / integration tests for this too.

lib/temporal/workflow/context.rb
```diff
@@ -123,27 +124,43 @@ def execute_workflow(workflow_class, *input, **args)
...
+        child_workflow_execution_future = Future.new(target, self, cancelation_id: cancelation_id)
+
+        child_workflow_future = ChildWorkflowFuture.new(
+          target, 
+          self, 
+          cancelation_id: cancelation_id, 
+          child_workflow_execution_future: child_workflow_execution_future
+        )
+
+        has_started_child_workflow_execution = false
```
antstorm, line 138: This is unused, right? ^

lib/temporal/workflow/context.rb
```diff
@@ -123,27 +124,43 @@ def execute_workflow(workflow_class, *input, **args)
         )
 
         target, cancelation_id = schedule_command(command)
-        future = Future.new(target, self, cancelation_id: cancelation_id)
+
+        # child_workflow_execution_future tracks when the workflow execution starts
+        child_workflow_execution_future = Future.new(target, self, cancelation_id: cancelation_id)
+
+        child_workflow_future = ChildWorkflowFuture.new(
```
antstorm, line 131: nit: you kinda already pass all the information need to create `child_workflow_execution_future` into the `ChildWorkflowFuture`, so you might avoid changing the initializer's signature and create the `Future` from within the `ChildWorkflowFuture`

antstorm (approved):
Couple nits, but otherwise looks great!

lib/temporal/workflow/context.rb
calum-stripe: nice idea, done!

lib/temporal/workflow/context.rb
calum-stripe: yes! sorry was using this inside the `error` handler to check if the future needed to be failed or not (then changed to how it is now with the current line)

calum-stripe:
@DeRauk check this out please!

DeRauk (approved):

PR #185, opened 2022-06-11
Allow clients to configure gRPC client interceptors

christopherb-stripe:
gRPC interceptors can be useful for adding common behavior and setting metadata for any interim service routing or proxies. This PR allows for clients to configure interceptors that will be executed as part of the request.

I'm not sold on the configuration for this as it will only apply to gRPC interceptors but it currently lives in the global configuration. Do you have a preference on how to expose connection type specific configuration?

I've added an integration test that shows the interface and provides a simple example.

What do you think?


christopherb-stripe:
Hmm, the integration test passes consistently locally on my machine. I'm not sure what's causing the failure here. I thought it could be that the client was being memoized without interceptors somewhere in the library but I would expect that to reproduce locally too.

PR #188, opened 2022-07-18
Allow initial search attributes to be specified when starting workflows

jazevedo-stripe:
### Summary
This PR allows initial search attributes to be specified when starting workflows, including the various different ways of doing so (normal start, start with signal, continue as new, child workflows). It passes the search attributes to the constructor of `ExecutionOptions`, which performs the common processing (ensure it is a hash, isn't nil, convert date-times to ISO strings) consistent with what the `upsert_search_attributes` method does.

I am new to temporal-ruby (and Ruby in general), so I based the changes in this PR on prior work in https://github.com/coinbase/temporal-ruby/pull/121 and https://github.com/coinbase/temporal-ruby/pull/145.

### Testing
- Changes to the unit tests based on the changes in https://github.com/coinbase/temporal-ruby/pull/121.
- Round-trip integration test in `examples/spec/integration/initial_search_attributes_spec.rb`

### DIFF
--- a/examples/spec/integration/initial_search_attributes_spec.rb
+++ b/examples/spec/integration/initial_search_attributes_spec.rb
@@ -1,28 +1,32 @@
 require 'workflows/upsert_search_attributes_workflow'
 require 'time'
 
-describe 'Temporal::Workflow::Context.upsert_search_attributes', :integration do
-  it 'can upsert a search attribute and then retrieve it' do
-    workflow_id = 'upsert_search_attributes_test_wf-' + SecureRandom.uuid
+describe 'Temporal::Client.start_workflow', :integration do
+  it 'can start a workflow with initial search attributes' do
+    workflow_id = 'initial_search_attributes_test_wf-' + SecureRandom.uuid
     expected_binary_checksum = `git show HEAD -s --format=%H`.strip
 
-    expected_added_attributes = {
-      'CustomStringField' => 'moo',
-      'CustomBoolField' => true,
-      'CustomDoubleField' => 3.14,
-      'CustomIntField' => 0,
+    initial_search_attributes = {
+      'CustomBoolField' => false,
+      'CustomIntField' => -1,
       'CustomDatetimeField' => Time.now.utc.iso8601,
+      # These should get overriden when the workflow upserts them
+      'CustomStringField' => 'meow',
+      'CustomDoubleField' => 6.28,
     }
+    upserted_search_attributes = {
+      'CustomStringField' => 'moo',
+      'CustomDoubleField' => 3.14,
+    }
+    expected_custom_attributes = initial_search_attributes.merge(upserted_search_attributes)
 
     run_id = Temporal.start_workflow(
       UpsertSearchAttributesWorkflow,
-      string_value: expected_added_attributes['CustomStringField'],
-      bool_value: expected_added_attributes['CustomBoolField'],
-      float_value: expected_added_attributes['CustomDoubleField'],
-      int_value: expected_added_attributes['CustomIntField'],
-      time_value: expected_added_attributes['CustomDatetimeField'],
+      string_value: upserted_search_attributes['CustomStringField'],
+      float_value: upserted_search_attributes['CustomDoubleField'],
       options: {
         workflow_id: workflow_id,
+        search_attributes: initial_search_attributes,
       },
     )
 
@@ -31,13 +35,13 @@ describe 'Temporal::Workflow::Context.upsert_search_attributes', :integration do
       workflow_id: workflow_id,
       run_id: run_id,
     )
-    expect(added_attributes).to eq(expected_added_attributes)
+    expect(added_attributes).to eq(upserted_search_attributes)
 
     # These attributes are set for the worker in bin/worker
     expected_attributes = {
       # Contains a list of all binary checksums seen for this workflow execution
       'BinaryChecksums' => [expected_binary_checksum]
-    }.merge(expected_added_attributes)
+    }.merge(expected_custom_attributes)
 
     execution_info = Temporal.fetch_workflow_execution_info(
       integration_spec_namespace,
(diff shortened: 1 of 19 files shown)


spec/unit/lib/temporal/grpc_spec.rb
```diff
@@ -39,6 +39,7 @@ class TestDeserializer
           run_timeout: 0,
           task_timeout: 0,
           memo: {},
+          search_attributes: {},
```
antstorm, line 42: It would be great to have at least one test example here that actually checks that search attributes are getting serialized properly

examples/spec/integration/initial_search_attributes_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+    upserted_search_attributes = {
+      'CustomStringField' => 'moo',
+      'CustomDoubleField' => 3.14,
+    }
+    expected_custom_attributes = initial_search_attributes.merge(upserted_search_attributes)
+
+    run_id = Temporal.start_workflow(
+      UpsertSearchAttributesWorkflow,
+      string_value: upserted_search_attributes['CustomStringField'],
+      float_value: upserted_search_attributes['CustomDoubleField'],
```
antstorm, line 26: I wonder if not passing bool and int fields here is intentional? Looks like you already have them in the `initial_search_attributes`, so probably missed here?

antstorm (changes requested):
Thank you for the PR, @jazevedo-stripe 👍 Left a few comments here and there, but overall looks good

lib/temporal/execution_options.rb
```diff
@@ -14,6 +15,7 @@ def initialize(object, options, defaults = nil)
       @timeouts = options[:timeouts] || {}
       @headers = options[:headers] || {}
       @memo = options[:memo] || {}
+      @search_attributes = Temporal::Workflow::Context::Helpers.process_search_attributes(options[:search_attributes] || {}, allow_empty: true)
 
       # For Temporal::Workflow and Temporal::Activity use defined values as the next option
       if has_executable_concern?(object)
```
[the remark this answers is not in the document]
jazevedo-stripe: I created the ability to specify default search attributes on the global `Temporal::Configuration` class, but I wasn't sure if I should also add the default to the Workflow class. This is because (if I understand it correctly) Workflow-class-defaults for the other properties (namespace, task queue, retry policy, timeouts, & headers) is done through the `Concerns::Executable` module/mixin, which is brought in for both activities and workflows (so adding a new option there for search attributes probably isn't appropriate).

Is this still something that should be implemented in this PR? And if so, how should it be done?

examples/spec/integration/initial_search_attributes_spec.rb
```diff
@@ -0,0 +1,63 @@
+require 'workflows/upsert_search_attributes_workflow'
+require 'time'
+
+describe 'Temporal::Client.start_workflow', :integration do
```
antstorm, line 4: nit: A relevant description probably makes more sense here, such as "Starting workflow with initial search attributes"

lib/temporal/execution_options.rb
```diff
@@ -1,9 +1,10 @@
 require 'temporal/concerns/executable'
 require 'temporal/retry_policy'
+require 'temporal/workflow/context_helpers'
```
antstorm, line 3: This is no longer a dependency of this class as far as I can tell

lib/temporal/execution_options.rb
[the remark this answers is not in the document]
antstorm: Happy to postpone this change to a follow up PR to get this one out of the way first. As for the specific implementation — you're right that `Concerns::Executable` won't work here, but you can simply put it in the `Workflow` class since it is a behaviour specific to it

lib/temporal/testing/temporal_override.rb
```diff
@@ -83,6 +83,7 @@ def start_locally(workflow, schedule, *input, **args)
         workflow_id = options[:workflow_id] || SecureRandom.uuid
         run_id = SecureRandom.uuid
         memo = options[:memo] || {}
+        initial_search_attributes = Workflow::Context::Helpers.process_search_attributes(options[:search_attributes] || {})
```
antstorm, line 86: I think you also need to `require` this module

lib/temporal/client.rb
```diff
@@ -61,6 +62,7 @@ def start_workflow(workflow, *input, options: {}, **args)
           workflow_id_reuse_policy: options[:workflow_id_reuse_policy],
           headers: execution_options.headers,
           memo: execution_options.memo,
+          search_attributes: Workflow::Context::Helpers.process_search_attributes(execution_options.search_attributes),
```
antstorm, line 65: Also missing a `require` statement for this helper

examples/spec/integration/initial_search_attributes_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+        search_attributes: initial_search_attributes,
+      },
+    )
+
+    added_attributes = Temporal.await_workflow_result(
+      UpsertSearchAttributesWorkflow,
+      workflow_id: workflow_id,
+      run_id: run_id,
+    )
+    expect(added_attributes).to eq(upserted_search_attributes)
```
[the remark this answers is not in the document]
antstorm: Ah, I see, sorry. My bad, got confused a bit there

lib/temporal/execution_options.rb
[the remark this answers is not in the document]
jazeved0: Ok, let's leave that to a follow up PR, then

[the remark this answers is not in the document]
jazevedo-stripe: Ok, let's leave that to a follow-up PR, then

antstorm (approved):
Looks great, thank you for contributing. Awesome job! 🙌

PR #190, opened 2022-08-17
Add an interface and support for cron scheduling child workflows

dwillett:
Currently there is an SDK method for `schedule_workflow`, but no corresponding ability to schedule a child workflow. This adds a corresponding helper method and a context method to execute a child workflow with a cron schedule.

Note: Is there any use case to support adding `schedule!`? 🤔


examples/spec/integration/schedule_child_workflow_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      }
+    )
+
+    # Giving the parent workflow sufficient time to run and schedule the child
+    sleep 2
+
+    history_response = fetch_history(child_workflow_id, nil)
+    expect(
+      history_response.history.events.first.workflow_execution_started_event_attributes.cron_schedule
+    ).to eq(cron_schedule)
```
dwillett, line 24: Not sure if there's a better way to prepare or assert this - maybe by fetching history for the parent workflow, if we're able to wait for a specific child workflow started event?

examples/spec/integration/schedule_child_workflow_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'workflows/schedule_child_workflow'
+require 'workflows/hello_world_workflow'
+
+describe ScheduleChildWorkflow, :integration do
+  let(:cron_schedule) { "*/6 * * * *" }
+
+  it 'schedules a child workflow with a given cron schedule' do
```
antstorm, line 7: We might also need to clean up the scheduled workflow after running this spec

examples/spec/integration/schedule_child_workflow_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      described_class,
+      child_workflow_id,
+      cron_schedule,
+      options: {
+        timeouts: { execution: 10 }
+      }
+    )
+
+    # Giving the parent workflow sufficient time to run and schedule the child
+    sleep 2
```
antstorm, line 19: Can we just wait for the completion here instead and only then check that the child workflow got scheduled? There should be some API that will allow to get details about a scheduled workflow before it is first executed. At least that's my guess…

lib/temporal/workflow/context.rb
```diff
@@ -130,6 +130,7 @@ def execute_workflow(workflow_class, *input, **args)
           parent_close_policy: parent_close_policy,
           timeouts: execution_options.timeouts,
           headers: execution_options.headers,
+          cron_schedule: execution_options.cron_schedule,
```
antstorm, line 133: We should handle it similarly to `parent_close_policy` and not add it to `execution_options` since these were meant for overlapping properties between Activities and Workflows. We might consider subclassing these later to host all options, but it feels like a separate refactor

lib/temporal/workflow/convenience_methods.rb
```diff
@@ -29,6 +29,13 @@ def execute!(*input, **args)
 
         context.execute_workflow!(self, *input, **args)
       end
+
+      def schedule(cron_schedule, *input, **args)
```
antstorm, line 33: This method is missing a spec

lib/temporal/workflow/context.rb
```diff
@@ -173,6 +174,11 @@ def execute_workflow!(workflow_class, *input, **args)
         result
       end
 
+      def schedule_workflow(workflow_class, cron_schedule, *input, **args)
```
antstorm, line 177: This is also missing a spec

examples/workflows/schedule_child_workflow.rb
```diff
@@ -0,0 +1,15 @@
+class ScheduleChildWorkflow < Temporal::Workflow
+  def execute(child_workflow_id, cron_schedule)
+    @state = "started"
```
antstorm, line 3: nit: no reason for an instance variable, can just be a regular variable

antstorm (changes requested):
Great PR, @dwillett. A few comments, but nothing too major

examples/spec/integration/schedule_child_workflow_spec.rb
antstorm: `DescribeWorkflowExecution`, which sits behind our `fetch_workflow_execution_info` and might do the trick…

examples/spec/integration/schedule_child_workflow_spec.rb
dwillett: It should be getting cleaned up currently, with the combination of the parent close policy and the execution timeout of the parent. Do you think we should further ensure the child is terminated?

examples/spec/integration/schedule_child_workflow_spec.rb
dwillett: I couldn't find anything related to the cron schedule surfaced in DescribeWorkflowExecution, but I was able to find it in the event history in my revised approach. Let me know what you think.

antstorm (approved):
Looks great 👍 Thank you!

PR #195, opened 2022-10-05
Fix for incorrect handling of Child Workflows that time out and are terminated

calum-stripe:
# Summary
This PR fixes two issues with children workflows:
1. A bug where when a child workflow times out, the method that generates the error fails because the sdk assumes the history return object has a `.failure` function / value and it doesn't. 
2. Terminated children workflows were never reported to the parent workflow (there was a previous unfinished todo around this in the `state_manager.rb` which is now completed

In both cases, we mark the child workflows future as failed with a corresponding error (`ChildWorkflowTimeoutError` or `ChildWorkflowTerminatedError`)

Also this introduces two integration workflows and tests that test that the above two behaviors are now supported. 

# Test Plan
```
# unit test
bundle exec rspec spec/

# integration test
cd examples
(terminal 1) ./bin/worker
```


examples/spec/integration/child_workflow_terminated_workflow_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      subject,
+      options: { workflow_id: workflow_id }
+    )
+
+    result = Temporal.await_workflow_result(
+      subject,
+      workflow_id: workflow_id
+    )
+
+    puts result
```
antstorm, line 19: Looks like this is a leftover from debugging the spec. It also seems to be missing an expectation

antstorm: I guess it's supposed to be `expect(result).to eq(child_workflow_terminated: true)`

lib/temporal/workflow/errors.rb
```diff
@@ -67,12 +67,19 @@ def self.generate_error_for_child_workflow_start(cause, workflow_id)
         end
       end
 
+      def self.generate_error_for_child_workflow_timeout
+        ChildWorkflowTimeoutError.new('The child workflow timed out before succeeding')
+      end
+
+      def self.generate_error_for_child_workflow_terminated
+        ChildWorkflowTerminatedError.new('The child workflow was terminated')
+      end
```
antstorm, line 76: nit: these 2 methods provide very little value that might not be worth the added indirection

examples/workflows/child_workflow_terminated_workflow.rb
```diff
@@ -0,0 +1,20 @@
...
+    child_workflow_execution = result.child_workflow_execution_future.get
+    TerminateWorkflowActivity.execute!(
+      'ruby-samples',
+      child_workflow_execution.workflow_id,
+      child_workflow_execution.run_id
+    )
+
+    # check that the result is now 'failed'
+    {
+      child_workflow_terminated: result.failed? # terminated is represented as failed? with the Terminated Error
```
antstorm, line 17: It would be good to return the actual error as well here, otherwise the spec won't fail if the child workflow fails for any other reason

examples/workflows/child_workflow_timeout_workflow.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class ChildWorkflowTimeoutWorkflow < Temporal::Workflow
+  def execute
+    # workflow timesout before it can finish running, we should be able to detect that with .failed?
+    result = QuickTimeoutWorkflow.execute
+
+    result.get # wait for the workflow to finish so we can detect if it failed or not
+
+    {
+      child_workflow_failed: result.failed?
```
antstorm, line 11: same here

antstorm (changes requested):
A few comments, but overall the solution looks correct and mostly ready

examples/workflows/child_workflow_terminated_workflow.rb
calum-stripe: agreed! added this to both places

examples/spec/integration/child_workflow_terminated_workflow_spec.rb
calum-stripe: agreed! added this to both places

lib/temporal/workflow/errors.rb
calum-stripe: dont disagree but just followed the same pattern to what we have today! What do you think?

antstorm: I don't think there's much benefit in doing this for every single error type, so we can probably break the patter for these simpler use-cases

calum-stripe: Let me update! :)

antstorm (approved):
The changes look good, but there are a few unit tests failing after the last commit. Not sure what specifically caused it, but seems consistent

antstorm (approved):
LGTM 🙌

PR #203, opened 2022-11-01
Count Workflow Executions

calum-stripe:


lib/temporal.rb
```diff
@@ -29,7 +29,8 @@ module Temporal
                  :fail_activity,
                  :list_open_workflow_executions,
                  :list_closed_workflow_executions,
-                 :query_workflow_executions
+                 :query_workflow_executions,
+                 count_workflow_executions
```
antstorm, line 33: This looks like a syntax error 👁️

lib/temporal/client.rb
```diff
@@ -409,6 +409,10 @@ def query_workflow_executions(namespace, query, next_page_token: nil, max_page_s
       Temporal::Workflow::Executions.new(connection: connection, status: :all, request_options: { namespace: namespace, query: query, next_page_token: next_page_token, max_page_size: max_page_size }.merge(filter))
     end
 
+    def count_workflow_executions(namespace, query)
```
antstorm, line 412: I don't think that other SDKs actually have this method exposed. With that in mind, let's not add it to the `Client`, but instead only expose it on the connection itself and possibly make the connection public. This way you can access the low-level API if needed, but we're not committing to supporting it via the high-level classes

antstorm (changes requested):

PR #207, opened 2022-12-08
Add task queue tag to workflow task failure metrics & task queue time metrics

bryton-stripe:
This PR emits the task queue as a tag for the following metrics:
- workflow task failures
- activity task queue times
- workflow task queue times

The existing tests for these metrics have been updated to expect the task queue to be emitted.

### DIFF
--- a/lib/temporal.rb
+++ b/lib/temporal.rb
@@ -29,7 +29,8 @@ module Temporal
                  :fail_activity,
                  :list_open_workflow_executions,
                  :list_closed_workflow_executions,
-                 :query_workflow_executions
+                 :query_workflow_executions,
+                 :connection
 
   class << self
     def configure(&block)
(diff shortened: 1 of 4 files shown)


antstorm (changes requested):
While the change itself makes sense, it's not going to work as expected. Think of `config` as a default set of option fallbacks, however these can be overriden per workflow or activity class:

```ruby
class MyActivity < Temporal::Activity
  task_queue 'my-task-queue'
end
```

In which case it will not get propagated correctly. You need to pass the `task_queue` from a poller explicitly to do this.

bryton-stripe:
Okay. workflow/activity-set task queues seem to take highest precedence as per [worker.rb](https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/worker.rb#L69), so this makes sense.

Does `Temporal.configuration` override the `config` defaults, then? That would explain the test behavior, since I manually set the `Temporal.configuration` and thus don't get back the default task queue in emitted metrics.

antstorm:
@bryton-stripe yes, that's exactly right (see https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/execution_options.rb for specific implementation).

In this case you can just pass in the `task_queue` from the `Poller`, which already has the correct one (that it's using to poll for activity/workflow tasks).

PR #208, opened 2022-12-13
dynamic Activity

drewhoskins-stripe:
Emulates the [DynamicActivity concept from the java SDK](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/activity/DynamicActivity.java).  [Javadocs](https://www.javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/activity/DynamicActivity.html)

This will allow folks to provide activity implementations on classes that are not Activities, an important extensibility mechanism.

Test Plan:
Unit: 
```
rspec ./spec/unit/lib/temporal/executable_lookup_spec.rb
```
Integration, using a sample:
Terminal 1:
```
cd examples
./bin/workers
```
Terminal 2:
```
cd examples
rspec ./spec/integration/dynamic_activity_spec.rb
```

Note logs name activities as `Plus` and `Times`, not `DelegatorActivity`


examples/activities/delegator_activity.rb
```diff
@@ -0,0 +1,35 @@
...
+    args[:a] * args[:b]
+  end
+end
+
+# Calls into our other class hierarchy.
+class DelegatorActivity < Temporal::Activity
+  dynamic
+
+  def execute(input)
+    executor = Object.const_get(activity.name).new
```
antstorm, line 30: I think it's a bit confusing that `activity.name` here is different to the activity itself. I would prefer the name of the "sub-activity" to be passed in as the first argument to this activity. I almost view this activity as Ruby's [`method_missing`](https://ruby-doc.org/core-2.6.3/BasicObject.html#method-i-method_missing)

antstorm (changes requested):
Thank you for the PR, @drewhoskins-stripe 🙌 

Overall looks like heading in the right direction, however I've outlined a few parts that we need to think a bit more about. Mostly I'm worried that we're trying to shove this into the current model, which wasn't really meant for it.

I think we can improve the API of how it is registered and then interacted with. I also wonder if we might want to take it a step further and allow specifying which exact sub-activities are there available on the dynamic activity.

Otherwise the only real difference from calling a regular activity (with a class name as the first arg) is that it tracks a different name in the history…

examples/activities/delegator_activity.rb
drewhoskins-stripe: added some comments and undid some unnecessary changes to the `Activity::Context` constructor (the name's already in the metadata)

antstorm (commented):
I think this makes sense, however since we're not exposing the "sub-activity" class name as a separate argument there's really no difference between a regular activity and a dynamic activity apart from the way they are registered with the worker (which I think works well).

With this in mind I think we should loose the `dynamic`/`dynamic?` methods from the `Temporal::Activity` class altogether and tell the lookup directly that we're registering a fallback vs a regular activity (rather than asking if it's `dynamic?` all the time). I think this will simplify the code quite a bit further and will remove the confusion around registering a dynamic activity as regular one and vice versa.

drewhoskins-stripe:
That's very reasonable and more elegant.  This will diverge us from the Java SDK, but I'm not wedded to that.

antstorm (approved):
Looks great, thank you for working on it! 🙌

DeRauk:
Thanks @drewhoskins-stripe!

DeRauk (approved):

PR #212, opened 2023-01-13
Allow exceptions thrown from activities to accept more args

drewhoskins-stripe:
# Description
Today, there is no way to throw an exception from an activity that can be handled by the workflow _unless_ it has exactly one args (and not a keyword args).

While there doesn't seem to be a general-purpose way to serialize and deserialize exceptions in ruby, we can allow the user to provide a serializer and deserializer.  We use a private class called `Temporal::Activity::SerializedException` to marshal the needed data.

# Test Plan
Unit tests for serialization and marshalling
`rspec ./spec/unit/lib/temporal/activity/serialized_exception_spec.rb`
`rspec ./spec/unit/lib/temporal/workflow/errors_spec.rb`
Integration test to make sure client and server work harmoniously:
```
cd examples
Terminal 1
./bin/worker
Terminal 2
rpsec spec/integration/handling_structured_error_workflow_spec.rb
```


lib/temporal/workflow/errors.rb
```diff
@@ -12,31 +13,42 @@ def self.generate_error(failure, default_exception_class = StandardError)
...
             exception_class = default_exception_class
           end
-
-
           begin
-            exception = exception_class.new(message)
-          rescue => deserialization_error
-            # We don't currently support serializing/deserializing exceptions with more than one argument.
-            message = "#{exception_class}: #{message}"
+            exception = if exception_class == Temporal::Activity::SerializedException
```
drewhoskins-stripe, line 28: This is slightly spaghetti-ish.  I can probably streamline it a bit after I've slept, but I'm keen to get your thumbs-up/thumbs-down on the overall PR, approach/UX first before spending too much time polishing details.

antstorm (commented):
Thank you for the PR, Drew, this is a very important topic that we kinda skimmed through so far.

Your implementation makes perfect sense, however we already started looking into this for the new SDK being built and had a few thoughts:

- Overall everyone's very reluctant to introduce a custom serialization format that's not based on any of the language's default patterns
- Based on this, we've decided to allow multiple options
- First of all, Ruby's `#dump` / `.load` marshalling interface for a fully-custom serialization (https://github.com/temporalio/proposals/blob/master/ruby/type-hinting-converters-codecs-errors.md#type-hinting, 2nd example)
- And also support JSON's Custom Additions (https://github.com/temporalio/proposals/blob/master/ruby/type-hinting-converters-codecs-errors.md#json-converter) to better fit into Ruby's ecosystem

Now, I know that these decisions were made for non-failure payloads (we've decided not to allow custom serialization for failures for now, waiting for a cross-language standard), but I feel like following Ruby's JSON Custom Additions makes the most sense here and very likely to be future-compatible. I also believe that `oj` supports that as well.

Hope this makes sense, but I'm happy to have a further discussion on it to ensure we're on the same page.

drewhoskins-stripe:
> And also support JSON's Custom Additions
Note that these are not supported in ruby 2.x (which is unfortunately where Stripe is right now).

PR #213, opened 2023-01-19
Paginate through empty workflow history pages

drewhoskins-stripe:
# Description
Temporal server sometimes sends empty history pages but with a next_page_token.  Best practice across the SDKs is to keep paginating.

But temporal ruby was frequently returning a `Temporal::UnexpectedResponse`. This caused workflow tasks to fail and delay ten seconds until the next retry.  Especially prevalent on long workflow histories.

This change simply removes the check since this is supported behavior.

Update: I threw in a couple of unrelated test fixes to get the CircleCI tests to pass.

# Test plan

Note : I OKed this with Chad Retz from Temporal.  We've had this change active for months at Stripe and have been happy with it.

New test:
`rspec spec/unit/lib/temporal/workflow/task_processor_spec.rb`

### DIFF
--- a/examples/bin/register_namespace
+++ b/examples/bin/register_namespace
@@ -4,7 +4,7 @@ require_relative '../init'
 namespace = ARGV[0]
 description = ARGV[1]
 
-fail 'Missing namespace name, please run register_namespace <namespace_name>' unless namespace
+raise 'Missing namespace name, please run register_namespace <namespace_name>' unless namespace
 
 begin
   Temporal.register_namespace(namespace, description)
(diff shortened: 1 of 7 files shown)


examples/spec/integration/search_attributes_spec.rb
```diff
@@ -47,24 +47,6 @@ def cleanup
     end.to raise_error(Temporal::SearchAttributeAlreadyExistsFailure)
   end
 
-  it 'type change fails' do
```
drewhoskins-stripe, line 50: @jeffschoner-stripe, this worked fine in the circleci version and I figure it's not really important functionality to test-- like if temporal fixes it, everyone will be happy--so I just removed the test.

examples/bin/register_namespace
```diff
@@ -13,4 +13,17 @@ rescue Temporal::NamespaceAlreadyExistsFailure
   Temporal.logger.info 'Namespace already exists', { namespace: namespace }
 end
 
-
+# Register a variety of search attributes for ease of integration testing
```
drewhoskins-stripe, line 16: fixes some tests that were broken in CI, I think based on some recent temporal docker image changes to stop registering these automatically.

jeffschoner-stripe: I think the behavior became inconsistent between how Temporal server is set up in temporalio/auto-setup (which the CI config here uses) and how it is setup in production and the docker-compose. The below code should handle both correctly, but Temporal should restore these to parity. I believe the auto-setup may be using SQL-based enhanced visibility where search attributes are scoped to a namespace and not global like with Elasticsearch.

examples/spec/integration/search_attributes_spec.rb
jeffschoner-stripe: I think this is another difference between SQL-based enhanced visibility and Elasticsearch-based visibility. I'm fine removing it for now, as we're mostly testing Temporal server, not temporal-ruby here.

jeffschoner-stripe (commented):
@drewhoskins-stripe This looks good me. Did you want to fix the error where the Temporal Java test service returns a slightly different proto representation for an incomplete workflow history when awaiting results? It's tangentially related to this change, but I can also do it in a separate PR at some point.

DeRauk (approved):

Issue #215, opened 2023-02-08
Log flood on connection error


jlemesh:
I’ve noticed that if I start Temporal Ruby worker and it is unable to connect to Temporal server, it logs big amount of errors. For example we get ~10k of log entries when Temporal is down for one second. I'm thinking of just doing some sleep before trying again, which could be configurable (default 0 to keep old behaviour and some configs for worker to specify how long would you like your pollers to sleep after unsuccessful poll). This way it will not only produce less logs but also won't bomb infrastructure with so many requests. I'll implement it if you're ok with this solution.

PR #216, opened 2023-02-08
Add sleep after bad poll

jlemesh:
Closes https://github.com/coinbase/temporal-ruby/issues/215

Add sleep before retrying after unsuccessful poll for workflow and activity tasks. Default value is 0 so without extra configs worker will behave in old way. Sleep duration is specified in seconds.
Also add tests for new functionality.
Add docs to `README.md`.

### DIFF
--- a/lib/temporal/activity/poller.rb
+++ b/lib/temporal/activity/poller.rb
@@ -91,7 +91,7 @@ module Temporal
 
         Temporal::ErrorHandler.handle(error, config)
 
-        sleep poll_retry_seconds unless poll_retry_seconds.zero?
+        sleep_before_retry(poll_retry_seconds) unless poll_retry_seconds.zero?
 
         nil
       end
@@ -115,6 +115,10 @@ module Temporal
 
       def poll_retry_seconds
         @options[:poll_retry_seconds]
+      end
+
+      def sleep_before_retry(seconds)
+        sleep seconds
       end
     end
   end
--- a/lib/temporal/activity/poller.rb
+++ b/lib/temporal/activity/poller.rb
@@ -91,7 +91,7 @@ module Temporal
 
         Temporal::ErrorHandler.handle(error, config)
 
-        sleep_before_retry(poll_retry_seconds) unless poll_retry_seconds.zero?
+        sleep(poll_retry_seconds)
 
         nil
       end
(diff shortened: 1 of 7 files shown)


lib/temporal/activity/poller.rb
```diff
@@ -90,6 +91,8 @@ def poll_for_task
 
         Temporal::ErrorHandler.handle(error, config)
 
+        sleep_before_retry(poll_retry_seconds) unless poll_retry_seconds.zero?
```
antstorm, line 94: nit: You don't really need the `poll_retry_seconds.zero?` check, as `sleep(0)` is a valid command and won't have any effect

spec/unit/lib/temporal/activity/poller_spec.rb
```diff
@@ -173,6 +174,45 @@ def call(_); end
...
+      allow(subject).to receive(:sleep_before_retry).and_return(nil)
+    end
+
+    it 'sleeps' do
+      subject.start
+
+      # stop poller before inspecting
+      subject.stop_polling; subject.wait
+
+      expect(subject).to have_received(:sleep_before_retry).with(5).once
```
antstorm, line 215: nit: you don't really need a `sleep_before_retry` as you can stub and expect `:sleep` to be called on the subject

spec/unit/lib/temporal/workflow/poller_spec.rb
```diff
@@ -178,6 +179,46 @@ def call(_); end
...
+        allow(subject).to receive(:sleep_before_retry).and_return(nil)
+      end
+
+      it 'sleeps' do
+        subject.start
+
+        # stop poller before inspecting
+        subject.stop_polling; subject.wait
+
+        expect(subject).to have_received(:sleep_before_retry).with(5).once
```
antstorm, line 220: Same here

antstorm (approved):
Couple of nits, but otherwise looking good 👍

jlemesh:
Fixed according to suggestions.

DeRauk (approved):

arrowcircle:
Looks like this does not work as expected.

```ruby
worker = Temporal::Worker.new(
  activity_poll_retry_seconds: 10
  workflow_poll_retry_seconds: 10
)
```

Does not have any effect and sends huge amount of errors in the logs.
Any ideas how to fix this?

PR #217, opened 2023-02-08
Update Temporal API to 1.16

jeffschoner:
### Summary

Updates the Temporal API to the most recent release. This hasn't been done in about 2 years.
* The proto submodule has been updated to the [1.16 release](https://github.com/temporalio/api/releases/tag/v1.16.0)
* The makefile for generating Ruby classes from proto has been updated to include the new `dependencies` top-level module that also needs to be generated
* All generated Ruby source files have been regenerated via the updated Makefile
* The proto definitions in the api now set a Ruby package of `Temporalio`. Previously, no package was set which caused the package to default to `Temporal`. Presumably, this is necessary for the new Rust-core based Temporal library. However, it also changes the fully qualified name of all generated classes. While this change affects many files, the rename is mechanical from `Temporal::Api` to `Temporalio::Api`.
* The `namespace` field has been removed from various payloads having to do with activity task scheduling. This removal has been plumbed through other parts of the library.

Reviewing commit-by-commit should show these changes more clearly.

### Testing
All existing unit and example specs succeed. The history_event_fabricator has been updated to remove the deprecated `namespace` field.


drewhoskins-stripe:
Should we avoid using `Temporalio` since the rust core SDK will likely use that?

jeffschoner:
> > Presumably, this is necessary for the new Rust-core based Temporal library.
> 
> Assuming we can configure it, should we therefore avoid using `Temporalio` and save it for the transition to the rust-based SDK?

@drewhoskins-stripe This comes from an annotation in the `.proto` files that applies to all Ruby code. You can see an example at https://github.com/temporalio/api/blob/master/temporal/api/operatorservice/v1/service.proto#L31, but this same annotation is in every Temporal API proto file now. Some quick searching around doesn't show a way to work around this. I'm not sure we would want to do such an override even if we could, because the top level package name is not declared in a single location.

drewhoskins-stripe:
Cool, thanks for clarifying!

DeRauk:
@jeffschoner Thanks for the PR!

I don't think we've typically done this, but since there are breaking changes could you bump the version to `0.0.2` in `lib/temporal/version.rb`?

jeffschoner:
> @jeffschoner Thanks for the PR!
> 
> I don't think we've typically done this, but since there are breaking changes could you bump the version to `0.0.2` in `lib/temporal/version.rb`?

Absolutely. It looks like there is also now a merge conflict with @drewhoskins-stripe's change that I'll need to resolve too.

jeffschoner:
@DeRauk I've rebased onto @drewhoskins-stripe's changes, and bumped the version number

DeRauk (approved):

jeffschoner:
That last push was just to clean up the author on my last commit. There's no code change.

antstorm (approved):
Looks great, thank you for the PR @jeffschoner  👍

PR #220, opened 2023-03-02
Terminate-if-running workflow ID reuse policy

jeffschoner:
### Summary

This new workflow ID reuse policy was added with the updated protobuf in #217. While the underlying GRPC connection layer already supports this with that updated proto, this change adds mapping code and tests for actually using it in temporal-ruby.

### Testing

New unit spec for mapper:
```
bundle exec rpsec spec/unit/lib/temporal/connection/serializer/workflow_id_reuse_policy_spec.rb
```

New example spec:
```
pushd examples
bin/worker # in separate terminal
bundle exec rspecspec/integration/start_workflow_spec.rb:73
popd
```

### DIFF
--- a/examples/spec/integration/start_workflow_spec.rb
+++ b/examples/spec/integration/start_workflow_spec.rb
@@ -1,4 +1,5 @@
 require 'workflows/hello_world_workflow'
+require 'workflows/long_workflow'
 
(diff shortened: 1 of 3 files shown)


DeRauk (approved):

antstorm (approved):
👍

PR #221, opened 2023-03-02
Drewhoskins schedule public

drewhoskins-stripe:
# Description

Emulates the [DynamicWorkflow](https://github.com/temporalio/sdk-java/blob/master/temporal-sdk/src/main/java/io/temporal/workflow/DynamicWorkflow.java) concept in the Java SDK.

This will allow folks to provide workflow implementations on classes that are not Workflows, an important extensibility mechanism. In particular, you can name your workflow whatever you want, so it will show up properly in Amp.

Using an identical technique to what we used for the [DynamicActivity PR](https://github.com/coinbase/temporal-ruby/pull/208).

# Test plan

New unit test: 
```
rspec spec/unit/lib/temporal/worker_spec.rb
```

New integration test:
```
# Console 1:
cd examples
./bin/worker
# Console 2:
rspec ./spec/integration/dynamic_workflow_spec.rb
```

### DIFF
--- a/examples/bin/worker
+++ b/examples/bin/worker
@@ -67,6 +67,7 @@ worker.register_workflow(UpsertSearchAttributesWorkflow)
 worker.register_workflow(WaitForWorkflow)
 worker.register_workflow(WaitForExternalSignalWorkflow)
 worker.register_workflow(WaitForNamedSignalWorkflow)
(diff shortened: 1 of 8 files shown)


DeRauk (approved):

DeRauk:
Thanks @drewhoskins-stripe!

PR #222, opened 2023-03-15
Read search attributes from inside workflow code

jeffschoner:
### Summary

Adds the ability to use `workflow.search_attributes` to read all current search attributes in workflow code

### Testing

New specs for state_manager:
```
bundle exec rspec spec/unit/lib/temporal/workflow/state_manager_spec.rb:42
```

Updated example specs for initial and upsert search attributes:
```
cd examples
bundle exec rspec spec/integration/initial_search_attributes_spec.rb spec/integration/upsert_search_attributes_spec.rb
```

### DIFF
--- a/examples/spec/integration/upsert_search_attributes_spec.rb
+++ b/examples/spec/integration/upsert_search_attributes_spec.rb
@@ -31,8 +31,7 @@ describe 'Temporal::Workflow::Context.upsert_search_attributes', :integration do
       workflow_id: workflow_id,
       run_id: run_id,
     )
-    expect(added_attributes[:from_code]).to eq(expected_added_attributes)
-    expect(added_attributes[:from_context]).to eq(expected_added_attributes)
+    expect(added_attributes).to eq(expected_added_attributes)
 
     # These attributes are set for the worker in bin/worker
     expected_attributes = {
--- a/examples/spec/integration/upsert_search_attributes_spec.rb
+++ b/examples/spec/integration/upsert_search_attributes_spec.rb
@@ -31,7 +31,8 @@ describe 'Temporal::Workflow::Context.upsert_search_attributes', :integration do
       workflow_id: workflow_id,
       run_id: run_id,
     )
-    expect(added_attributes).to eq(expected_added_attributes)
+    expect(added_attributes[:from_code]).to eq(expected_added_attributes)
+    expect(added_attributes[:from_context]).to eq(expected_added_attributes)
 
     # These attributes are set for the worker in bin/worker
     expected_attributes = {
--- a/lib/temporal/workflow/context.rb
+++ b/lib/temporal/workflow/context.rb
@@ -55,6 +55,10 @@ module Temporal
         metadata.headers
       end
 
+      def search_attributes
+        state_manager.search_attributes
+      end
+
       def has_release?(release_name)
         state_manager.release?(release_name.to_s)
       end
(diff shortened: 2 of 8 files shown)


lib/temporal/workflow/context.rb
```diff
@@ -55,6 +55,10 @@ def headers
         metadata.headers
       end
 
+      def search_attributes
```
drewhoskins-stripe, line 58: nit: you might want to add a comment telling people about the two ways to put things here.

lib/temporal/workflow/state_manager.rb
```diff
@@ -123,6 +124,10 @@ def apply_event(event)
 
         case event.type
         when 'WORKFLOW_EXECUTION_STARTED'
+          unless event.attributes.search_attributes.nil?
+            search_attributes.merge!(from_payload_map(event.attributes.search_attributes&.indexed_fields || {}))
```
drewhoskins-stripe, line 128: nit: You don't need the `&`, and possibly not the `|| {}` depending on the default for `indexed_fields`

spec/unit/lib/temporal/workflow/state_manager_spec.rb
```diff
@@ -36,4 +38,30 @@ class MyWorkflow < Temporal::Workflow; end
       end
     end
   end
-end
\ No newline at end of file
+
+  describe '#search_attributes' do
+    let(:start_workflow_execution_event) { Fabricate(:api_workflow_execution_started_event) }
```
drewhoskins-stripe, line 43: not sure how common it will be, but you should probably test a third, overwritten search attribute as well to demonstrate the upsert semantics.

spec/unit/lib/temporal/workflow/state_manager_spec.rb
```diff
@@ -36,4 +38,30 @@ class MyWorkflow < Temporal::Workflow; end
       end
     end
   end
-end
\ No newline at end of file
+
+  describe '#search_attributes' do
```
drewhoskins-stripe, line 42: Can you add a test when no search attributes are set to make sure it returns `{}` ?

drewhoskins-stripe (changes requested):

examples/spec/integration/upsert_search_attributes_spec.rb
```diff
@@ -31,7 +31,8 @@
       workflow_id: workflow_id,
       run_id: run_id,
     )
-    expect(added_attributes).to eq(expected_added_attributes)
+    expect(added_attributes[:from_code]).to eq(expected_added_attributes)
```
drewhoskins-stripe, line 34: nit: Appreciate keeping the existing test coverage in place, but it's probably not necessary--just drop the `:from_code` part and change the example?

lib/temporal/workflow/state_manager.rb
jeffschoner: I think this is necessary because Ruby GRPC treats set and unset properties differently: `{}` if set but `nil` if unset. This works differently than Java GRPC which always sets a default value (here, `{}`). I've seen some other places where this behavior crops up in temporal-ruby where it's a non-issue in the Java SDK.

Moreover, I've seen differences where the Temporal server will set the value but the test-server emulator in the Java SDK does not. Because of the Java SDK's behavior with defaults, the difference doesn't matter to it, but causes divergent behavior in temporal-ruby between real and test servers. Better to protect against these differences and even against breaking changes that could be made in Temporal server.

examples/spec/integration/upsert_search_attributes_spec.rb
jeffschoner: Ended up refactoring this a bit to make it more straightforward and useful in verifying correctness

jeffschoner:
@drewhoskins-stripe Please take another look

DeRauk (approved):

PR #224, opened 2023-03-21
Adding payload codec pipeline

calum-stripe:
# Overview
Adds the payload codec pipeline described here https://docs.temporal.io/security#payload-codec. 

This adds a new configuration option called `payload_codec` which always runs after the payload has been serialized (and before deserialization) when sent to (and received from) the server.

Currently, all payloads go through the data converter but only non search attributes payloads should go through payload codecs (otherwise if the payload is transformed into something the server can not read e.g. compressed or encrypted, then the server will reject it because it can not insert non datetime strings for example into the es index) 

# Test Plan
```
# unit test
bundle exec rspec spec/

# integration test
cd examples
(terminal 1) ./bin/worker
(terminal 3) USE_ERROR_SERIALIZATION_V2=1 ./bin/worker
(terminal 4) bundle exec rspec spec/integration
```


bergundy:
In the official SDKs search attributes do not go through the user configurable data converter, might want to do the same here.

calum-stripe:
>In the official SDKs search attributes do not go through the user configurable data converter, might want to do the same here.
this is true, I wonder if anyone has built a data converter for their search attributes though that by removing this and just using JSON would break their existing setup? Should we bump the version just incase?

lib/temporal/connection/converter/codec/base.rb
```diff
@@ -0,0 +1,33 @@
...
+            )
+          end
+
+          def decodes(payloads)
+            return nil if payloads.nil?
+
+            Temporalio::Api::Common::V1::Payloads.new(
+              payloads: payloads.payloads.map(&method(:decode))
+            )
+          end
```
jeffschoner-stripe, line 20: Would these be overridden by someone extending this class? Why not move these into `payloads.rb`?

lib/temporal/connection/converter/codec/chain.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module Temporal
+  module Connection
+    module Converter
+      module Codec
+        # Performs encoding/decoding on the payloads via the given payload codecs. When encoding
+        # the codecs are applied last to first meaning the earlier encodings wrap the later ones.
+        # When decoding, the codecs are applied first to last to reverse the effect.
+        class Chain < Base
+          def initialize(payload_codecs:)
```
jeffschoner-stripe, line 11: nit: This might be more ergonomic with a splat or non-keyword argument instead

spec/unit/lib/temporal/connection/converter/codec/base_spec.rb
```diff
@@ -0,0 +1,76 @@
+require 'temporal/connection/converter/codec/chain'
+
+describe Temporal::Connection::Converter::Codec::Base do
```
jeffschoner-stripe, line 3: Testing the base class seems strange to me since no one is expected to use this directly. It will always be extended, right? This might be a further indication that these non-abstract methods should simply be moved into `payloads.rb` instead.

jeffschoner-stripe (commented):
Is it possible to update the existing encryption test to use this new pipeline? I think it's better suited to your solution rather than switching out the converter like it does today.

lib/temporal/concerns/payloads.rb
```diff
@@ -62,6 +78,10 @@ def to_payload_map(data)
       def payload_converter
         Temporal.configuration.converter
       end
+
+      def payload_codec
```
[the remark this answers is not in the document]
calum-stripe: added to the `configuration.rb`, let me know what you think

lib/temporal/connection/converter/codec/base.rb
calum-stripe: they can be, I just was following the same pattern as in `payloads.rb` which seems to be define the same thing (`from_payloads` and `to_payloads` with logic then `from_payload` and `to_payload` are empty and required)

You probably are going to only override `encode` or `decode` but you have the option to override all for if the use case needs it.

lib/temporal/connection/converter/codec/chain.rb
calum-stripe: i agree, i was just keeping the same pattern as https://github.com/coinbase/temporal-ruby/blob/378630a545ac52759ac1030397f66aebfdec24df/lib/temporal/connection/converter/composite.rb#L10, what do you think?

spec/unit/lib/temporal/connection/converter/codec/base_spec.rb
calum-stripe: hmm, i don't disagree, just was following the pattern that was already in place for `from_payloads` and `to_payloads`. What do you think? Im easy either way.

calum-stripe: if we do remove, ill remove the unit tests

jeffschoner-stripe (approved):

DeRauk (approved):

PR #234, opened 2023-04-11
Heartbeat throttling

jeffschoner:
### Summary

In order to reduce load and potential denial of service of the Temporal server, the Go and Java SDKs limit the number of heartbeats sent by activities. This also has cost implications for Temporal Cloud which charges per heartbeat. This change adds this same kind of throttling here.

If no heartbeat has been sent yet, it will still be immediately sent, just like before this change. Any error or cancellation will be returned back to the caller. However, it will schedule a timer at the desired heartbeating interval on a background thread. While this timer is scheduled, any heartbeat payloads will be saved in memory but not actually sent to Temporal server. Once the interval has passed, any the latest saved heartbeat payload will be sent. If there is an unsent heartbeat when an activity completes, it will be discarded.

The timer functionality is facilitated by sleeping on a thread. To avoid the overhead of creating a new thread for each activity execution, a secondary heartbeat thread pool has been created of the same size as the main activity thread pool. A `ScheduledThreadPool` class is introduced for scheduling items on the thread pool threads. These items are cancelable.

### Testing

There are unit dedicated specs for the scheduled thread pool at,
```
bundle exec rspec spec/unit/lib/temporal/scheduled_thread_pool_spec.rb
```

Many other specs have updated or extended to cover this functionality,
```
bundle exec rspec \
  spec/unit/lib/temporal/activity/context_spec.rb \
  spec/unit/lib/temporal/activity/poller_spec.rb \
  spec/unit/lib/temporal/activity/task_processor_spec.rb \
  spec/unit/lib/temporal/metadata/activity_spec.rb \
  spec/unit/lib/temporal/activity/context_spec.rb
```

The existing example spec for activity cancellation still works without change, even though some of the underlying mechanisms for activity cancellation have changed:
```
cd examples # run 3 workers according to README.md
bundle exec rspec spec/integration/activity_cancellation_spec.rb
```

### DIFF
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -17,9 +17,10 @@ module Temporal
         @heartbeat_mutex = Mutex.new
         @async = false
         @cancel_requested = false
+        @last_heartbeat_throttled = false
       end
 
-      attr_reader :heartbeat_check_scheduled, :cancel_requested
+      attr_reader :heartbeat_check_scheduled, :cancel_requested, :last_heartbeat_throttled
 
       def async
         @async = true
@@ -69,10 +70,12 @@ module Temporal
           if heartbeat_check_scheduled.nil?
             send_heartbeat(details)
             @last_heartbeat_details = []
+            @last_heartbeat_throttled = false
             @heartbeat_check_scheduled = schedule_check_heartbeat(heartbeat_throttle_interval)
           else
             logger.debug('Throttling heartbeat for sending later', metadata.to_h)
             @last_heartbeat_details = [details]
+            @last_heartbeat_throttled = true
           end
         end
 
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -13,13 +13,13 @@ module Temporal
         @config = config
         @heartbeat_thread_pool = heartbeat_thread_pool
         @last_heartbeat_details = [] # an array to differentiate nil hearbeat from no heartbeat queued
-        @heartbeat_scheduled = nil
+        @heartbeat_check_scheduled = nil
         @heartbeat_mutex = Mutex.new
         @async = false
         @cancel_requested = false
       end
 
-      attr_reader :heartbeat_scheduled, :cancel_requested
+      attr_reader :heartbeat_check_scheduled, :cancel_requested
 
       def async
         @async = true
@@ -66,10 +66,10 @@ module Temporal
         # ...
 
         heartbeat_mutex.synchronize do
-          if heartbeat_scheduled.nil?
+          if heartbeat_check_scheduled.nil?
             send_heartbeat(details)
             @last_heartbeat_details = []
-            @heartbeat_scheduled = schedule_heartbeat(heartbeat_throttle_interval)
+            @heartbeat_check_scheduled = schedule_check_heartbeat(heartbeat_throttle_interval)
           else
             logger.debug('Throttling heartbeat for sending later', metadata.to_h)
             @last_heartbeat_details = [details]
@@ -144,14 +144,17 @@ module Temporal
         end
       end
 
-      def schedule_heartbeat(delay)
+      def schedule_check_heartbeat(delay)
         heartbeat_thread_pool.schedule([metadata.workflow_run_id, metadata.id, metadata.attempt], delay) do
           details = heartbeat_mutex.synchronize do
-            @heartbeat_scheduled = nil
+            @heartbeat_check_scheduled = nil
+            # Check to see if there is a saved heartbeat. If heartbeat was not called while this was waiting,
+            # this will be empty and there's no need to send anything or to scheduled another heartbeat
+            # check.
             last_heartbeat_details
           end
           begin
-            if !details.nil?
+            unless details.empty?
               heartbeat(details.first)
             end
           rescue
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -145,6 +145,8 @@ module Temporal
       end
 
       def schedule_check_heartbeat(delay)
+        return nil if delay <= 0
+
         heartbeat_thread_pool.schedule([metadata.workflow_run_id, metadata.id, metadata.attempt], delay) do
           details = heartbeat_mutex.synchronize do
             @heartbeat_check_scheduled = nil
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -117,6 +117,8 @@ module Temporal
       end
 
       def heartbeat_throttle_interval
+        # This is a port of logic in the Go SDK
+        # https://github.com/temporalio/sdk-go/blob/eaa3802876de77500164f80f378559c51d6bb0e2/internal/internal_task_handlers.go#L1990
         interval = if metadata.heartbeat_timeout > 0
           metadata.heartbeat_timeout * 0.8
         else
--- a/lib/temporal/activity/context.rb
+++ b/lib/temporal/activity/context.rb
@@ -7,11 +7,19 @@ require 'temporal/activity/async_token'
 module Temporal
   class Activity
     class Context
-      def initialize(connection, metadata)
+      def initialize(connection, metadata, config, heartbeat_thread_pool)
         @connection = connection
         @metadata = metadata
+        @config = config
+        @heartbeat_thread_pool = heartbeat_thread_pool
+        @last_heartbeat_details = [] # an array to differentiate nil hearbeat from no heartbeat queued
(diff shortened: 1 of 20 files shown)


lib/temporal/configuration.rb
```diff
@@ -31,7 +31,12 @@ class Configuration
       # See       # https://docs.temporal.io/blog/activity-timeouts/#schedule-to-start-timeout
       schedule_to_start: nil,
       start_to_close: 30,     # Time spent processing an activity
-      heartbeat: nil          # Max time between heartbeats (off by default)
+      heartbeat: nil,         # Max time between heartbeats (off by default)
+      # If a heartbeat timeout is specified, 80% of that value will be used for throttling. If not specified, this
+      # value will be used.
+      default_heartbeat_throttle_interval: 30,
```
calum-stripe, line 37: are these the same as Java / Go and if not, how did you arrive at these!

lib/temporal/configuration.rb
```diff
@@ -31,7 +31,12 @@ class Configuration
...
+      # Heartbeat throttling interval will always be capped by this value
+      max_heartbeat_throttle_interval: 60
```
calum-stripe, line 39: same question as above!

calum-stripe (approved):
Looks good, my only questions are around where we got the defaults from!

lib/temporal/configuration.rb
jeffschoner: These came from the Go SDK. I believe they are the same in Java.

lib/temporal/activity/context.rb
```diff
@@ -64,11 +94,54 @@ def name
...
+        rescue => error
+          Temporal::ErrorHandler.handle(error, config, metadata: metadata)
+          raise
+        end
+      end
+
+      def schedule_heartbeat(delay)
+        heartbeat_thread_pool.schedule([metadata.workflow_run_id, metadata.id, metadata.attempt], delay) do
+          details = heartbeat_mutex.synchronize do
+            @heartbeat_scheduled = nil
```
drewhoskins-stripe, line 132: Suppose a user is sending a heartbeat every 10ms and the heartbeat throttle interval is 4 seconds.
If I'm reading this right, we'll send heartbeats at 10ms, 4.010s, 4.020s, 8.020s, 8.030s, ... because we clear the scheduled heartbeat-- so the next time the user heartbeats, we'll  immediately send it.
Is there any reason not to schedule the next one immediately so that we end up sending half as many heartbeats?

In either case, probably adding a high-level example will help folks understand this code.

lib/temporal/activity/context.rb
```diff
@@ -30,9 +39,30 @@ def async_token
         )
       end
 
+      def timed_out?
+        !start_to_close_deadline.nil? && Time.now > start_to_close_deadline
+      end
+
       def heartbeat(details = nil)
```
drewhoskins-stripe, line 46: Not exactly related, but since you're adding `ActivityTimedOut` it may be an opportune scope creep... Another thing we want to do here is detect if the worker is draining, i.e. if `Temporal::Activity::Poller.shutting_down?` is true, and raise if it is.  (I'll send you a link to the JIRA internally)  One idea is to have the poller set a thread-local under a mutex that we can check here.

lib/temporal/configuration.rb
drewhoskins-stripe: Yeah, mentioning the provenance in the code would be great.

drewhoskins-stripe (commented):
should we add a rollout mechanism here?

jeffschoner:
> should we add a rollout mechanism here?

This could be done by adding a config option to toggle this behavior. I think this would be wise in case existing users are relying on the existing behavior.

lib/temporal/activity/context.rb
```diff
@@ -64,11 +94,54 @@ def name
...
+            logger.info('Activity has been canceled', metadata.to_h)
+            @cancel_requested = true
+          end
+        rescue => error
+          Temporal::ErrorHandler.handle(error, config, metadata: metadata)
+          raise
+        end
+      end
+
+      def schedule_heartbeat(delay)
```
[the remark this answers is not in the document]
jeffschoner: I'll clean this up. Before I fully reconciled the behavior with the Go SDK, this was working slightly differently and was actually scheduling. Now, it scheduling to _possibly_ send a heartbeat if another one was received which is slightly different.

lib/temporal/activity/context.rb
jeffschoner: I'll add an example of this and confirm the behavior. We do schedule the next one immediately because when the timer wakes up, it calls the function recursively (from the callback so it's not typical recursion).

Your example is tricky to explain precisely because there's a race condition between the scheduled heartbeat and the next one coming in. But here's roughly how it would work:

10ms: first heartbeat request
- gets sent immediately
- assuming that call takes 100ms to complete a timer will be scheduled to fire at 4110ms to send any future heartbeat

20ms: second heartbeat request
- value saved for later

... same thing happens again, last heartbeat value gets updated over and over

4100ms: 410th heartbeat comes in, value saved
4110ms: timer goes off (let's assume it wins the race condition against the 411th heartbeat request)
- since there is a saved heartbeat, we send it immediately, this is the 410th heartbeat's value
- assume sending the heartbeat takes 100ms again
- timer scheduled for 8210ms (4110 + 100 + 4000)

Also at 4110ms on another thread: this 411th heartbeat is blocked obtaining the mutex since a heartbeat is being sent  on the first thread until 4210ms (when the heartbeat finishes being sent)

4210ms: 411th heartbeat finally gets the mutex. Because another timer has been scheduled, its value saved and not sent immediately 

... same thing happens again

The scheduling chain would only be broken if there's no saved heartbeat when the timer wakes up (i.e., no heartbeat was sent during the entire throttling interval). In that case, it has no heartbeat to send, so therefore does not send it and does not recursively schedule a future heartbeat.

lib/temporal/activity/context.rb
jeffschoner: I was thinking a bit about this while coding this up. I was considering a more complicated solution originally involving raising a special error on each heartbeat thread but checking the centralized `shutting_down?` bit on the poller may be sufficient and simpler. I'll see how complicated this might be before deciding to include it in this PR or to break it out into a follow up.

lib/temporal/activity/context.rb
jeffschoner: Added a simpler, easier to follow example into a comment

jeffschoner:
@drewhoskins-stripe This should have all the changes you asked about above. I've removed the start-to-close timeout check. I'm going to pull this into a separate PR along with communicating worker shutdown to activities, and unify those with activity cancelation.

lib/temporal/scheduled_thread_pool.rb
```diff
@@ -0,0 +1,111 @@
+require 'temporal/metric_keys'
+
+# This class implements a thread pool for scheduling tasks with a delay.
+# If threads are all occupied when a task is scheduled, it will be queued
+# with the sleep delay adjusted based on the wait time. 
+module Temporal
+  class ScheduledThreadPool
```
DeRauk, line 7: @jeffschoner is this something you all have implemented internally and tested in other services?

jeffschoner: @DeRauk This has not run yet in production services yet. I have done testing in our staging environment. CI for all our tests pass as well. Instead of the `LocalWorkflowContext`-based unit testing mode, almost all of our workflow tests in CI use real workers against the Temporal emulator from the Java SDK. Our CI does also look for problems like thread leaks.

My plan is to deploy this next week with a max heartbeat throttle timeout of 0, which will use the new code path but effectively disable throttling. This setting is behind a feature flag that we'll use to roll out the 60s default max to all namespaces gradually. I'm glad to hold off on merging this until that process is done to have a higher level of confidence in the change.

jeffschoner:
Rebased onto master to get CI fixes

jeffschoner:
@DeRauk Rollout is still proceeding on this to production, but so far it is looking good. We have seen cases where this is reducing the number of heartbeats sent to Temporal server substantially. Beyond reducing load on the server, it also is improving the performance of these activities. In one activity, we saw that 95% of heartbeats now being throttled. That activity also a 30% reduction in duration without having all these synchronous heartbeat calls blocking progress. That ultimately translates into 30% fewer workers needed to handle the same load.

The only thing I have found so far was that we did want to be able to check if a heartbeat got throttled for internal instrumentation reasons. I've added a method for this flag to activity context.

DeRauk:
Thanks @jeffschoner! Those results are great, I'll get this merged.

DeRauk (approved):

PR #235, opened 2023-04-12
Github Actions CI

DeRauk:
**Summary**
This PR integrates the test suite with Github Actions and runs tests on PRs and master commits.

Our CircleCI integration was disabled recently due to a bad deployment key. Our security team asked us to use this opportunity to move to Github Actions instead of CircleCI. 

I had to make a few test changes to get things working in Github Actions.

**Changelist**
 - Remove CircleCI config and add Github Actions config
 - Remove Coveralls - This was broken in Github Actions. We can look at adding it back later if there's a need.
 - Remove dynamic config file path from examples temporal container. This was breaking the container and the dynamic config is empty anyways.
 - Add sleep to `ChildWorkflowTerminatedWorkflow` spec to reduce flakiness.
 - Mock out workflow and activity poller in worker start specs to stop zombie threads from spawning.
 - Kill specific worker pid instead of `0` in signal worker specs. These were killing the github actions process and failing the test suite when we were killing `0`


examples/workflows/child_workflow_terminated_workflow.rb
```diff
@@ -12,6 +12,9 @@ def execute
       child_workflow_execution.run_id
     )
 
+    # Give time for termination to propagate
+    workflow.sleep(1)
```
jeffschoner-stripe, line 16: I'm surprised this is necessary. I'd expect the parent workflow to not replay until the activity has returned and the child has actually been terminated. It may just need another workflow task to be scheduled (which gets caused by `.sleep`) before the history entry for the terminated child finds its way into the parent. I think this is fine for now, but possibly something to look into.

jeffschoner-stripe (approved):

examples/workflows/child_workflow_terminated_workflow.rb
DeRauk: That's possible. This was broken in CI and on my machine until I added a `workflow.logger.info` statement here. I switched it to a sleep as a guess that it just needed more time but didn't look into it in depth.

cj-cb (approved):

PR #240, opened 2023-05-02
Make worker.start and worker.stop threadsafe

jeffschoner:
Upstreamed on behalf of @drewhoskins-stripe. Merge conflicts with #216 resolved.

# Description
In tests, we need to stop the worker threads from the main thread when we tear down the test case.
This was racing with start if we called it too soon, as was common, for example, if we threw an exception when trying to create a workflow.
Most commonly this resulted in a deadlock:
* stop method begins, shuts down 0 pollers, and sleeps
* start method completes, adding more pollers as part of normal initialization.
* stop method now calls `:wait` on each poller, but since they were added after the shutdown, they had not previously been shut down.

There are other potentially subtle issues, so I'm just throwing start and stop under a mutex.
I also added an assertion to the pollers to catch when this deadlock would happen in the future.

# Test Plan
I needed to refactor a bunch of  fragile unit tests that depended on the exact sequence of calls to `:shutting_down?`.

```
bundle exec rspec -f d ./spec/unit//lib/temporal/worker_spec.rb
bundle exec rspec -f d ./spec/unit//lib/temporal/workflow/poller_spec.rb
bundle exec rspec -f d ./spec/unit//lib/temporal/activity/poller_spec.rb
```
plus integration tests
```
cd examples
./bin/worker&
rspec ./spec/integration
```

### DIFF
--- a/lib/temporal/worker.rb
+++ b/lib/temporal/worker.rb
@@ -45,6 +45,7 @@ module Temporal
         binary_checksum: binary_checksum,
         poll_retry_seconds: workflow_poll_retry_seconds
       }
+      @start_stop_mutex = Mutex.new
     end
 
     def register_workflow(workflow_class, options = {})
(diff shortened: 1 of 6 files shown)


jeffschoner:
@jlemesh This is a change @drewhoskins-stripe made a couple years ago, but it was never merged back to Coinbase master. Because resolving merge conflicts with your change #216 was somewhat tricky, I'd appreciate if you could give this a look.

jlemesh (approved):

calum-stripe (approved):
This looks good! Thanks @jeffschoner for bringing this up from the fork.

jeffschoner:
Cleaned up some unnecessary `poll` method parameters, and removed unnecessary `allow` statements in poller specs. These changes are all superficial.

lib/temporal/worker.rb
```diff
@@ -143,6 +153,10 @@ def shutting_down?
       @shutting_down
     end
 
+    def on_started_test_hook; end
+    def while_stopping_test_hook; end
+    def on_stopped_test_hook; end
```
DeRauk, line 158: @jeffschoner this is a nitpick but do you mind removing the `test` part of these hook names? So use `on_started_hook`, `while_stopping_hook`, and `on_stopped_hook`?

jeffschoner:
@DeRauk Renamed the hook methods

DeRauk (approved):

Issue #237, opened 2023-04-26
NameError: uninitialized constant Temporal::VERSION


bmorton:
I just updated to the latest SHA and noticed some recent additions for custom search attributes that I wanted to test out.  However, I've run into this exception while trying to call `Temporal.add_custom_search_attributes`.  In my codebase, if I add `require "temporal/version"` in an initializer, everything is good.  Do we need to do that somewhere in the gem?  Should it be required [here](https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal.rb)?  Happy to open a PR if it's helpful!

```
rake aborted!
NameError: uninitialized constant Temporal::VERSION
Did you mean?  REVISION
/opt/app/laddertruck/vendor/bundle/ruby/3.0.0/bundler/gems/temporal-ruby-59ed5911850e/lib/temporal/connection/interceptors/client_name_version_interceptor.rb:8:in `request_response'
/opt/app/laddertruck/vendor/bundle/ruby/3.0.0/gems/grpc-1.38.0/src/ruby/lib/grpc/generic/interceptors.rb:175:in `intercept!'
/opt/app/laddertruck/vendor/bundle/ruby/3.0.0/gems/grpc-1.38.0/src/ruby/lib/grpc/generic/client_stub.rb:179:in `request_response'
/opt/app/laddertruck/vendor/bundle/ruby/3.0.0/gems/grpc-1.38.0/src/ruby/lib/grpc/generic/service.rb:171:in `block (3 levels) in rpc_stub_class'
/opt/app/laddertruck/vendor/bundle/ruby/3.0.0/bundler/gems/temporal-ruby-59ed5911850e/lib/temporal/connection/grpc.rb:511:in `add_custom_search_attributes'
/opt/app/laddertruck/vendor/bundle/ruby/3.0.0/bundler/gems/temporal-ruby-59ed5911850e/lib/temporal/client.rb:415:in `add_custom_search_attributes'
```

PR #241, opened 2023-05-02
Require version in client name version interceptor

jeffschoner:
This fixes https://github.com/coinbase/temporal-ruby/issues/237 (an issue with package loading) introduced in #223.

fyi @mjameswh

### DIFF
--- a/lib/temporal/connection/interceptors/client_name_version_interceptor.rb
+++ b/lib/temporal/connection/interceptors/client_name_version_interceptor.rb
@@ -1,4 +1,5 @@
 require 'grpc'
+require 'temporal/version'
 
 module Temporal
   module Connection


mjameswh:
Oops. Not sure how I can have missed this, but that's right.

Regarding the errors in get in CI tests:

```
:search attribute CustomDatetimeField is not defined.
```

In some recent servers (including the Temporal CLI dev server), customer search attributes are no longer present by default. You need to create them yourself.

See this for an example: https://github.com/temporalio/sdk-typescript/blob/main/packages/test/src/helpers.ts#L132

mjameswh:
Oh... there's a cache on server regarding search attributes. That sometime cause a lag (up to 10 seconds) before newly created search attributes are visible to workflow operations.

A very simple solution is to add `
--dynamic-config-value system.forceSearchAttributesCacheRefreshOnRead=true
` add argument to the server.

jeffschoner:
@mjameswh We are using auto-setup, which I think does provision all of these ([source](https://github.com/temporalio/docker-builds/blob/main/docker/auto-setup.sh)). This would also seem to explain why in the Register namespace Github Action step, it mentions all the custom search attributes have been created except for `CustomStringAttribute`. What's odd there is that there's a comment to remove `CustomStringAttribute` eventually. It seems an odd coincidence that attribute would not have been provisioned. Is it possible the docker image contains newer code than what's on master in the link I provided?

Either way, it's not clear to me how to pass that dynamic config flag through to the docker-compose file. Is there an easy way to do this?

There is a sleep step in the Github Actions config that waits for 10s after our set up script runs. It seems like that could also be extended, but `CustomDatetimeField` seems to have been created maybe as much as 45s before it was attempted to be used.

DeRauk (approved):

DeRauk:
@jeffschoner I'm going to merge this since the test failures aren't related to this change and we need to get this in.

I put out a PR to add the setting @mjameswh mentioned here: https://github.com/coinbase/temporal-ruby/pull/242

Thank you both!

PR #247, opened 2023-06-02
Fix and add spec for local activities

### LINKED ISSUE #245: ArgumentError: wrong number of arguments (given 2, expected 4)
#234 introduced a bug here: https://github.com/coinbase/temporal-ruby/blob/41a8e923e52f62a6727e3ce6ffa2d68090b29439/lib/temporal/workflow/context.rb#L116

We introduced 2 new arguments to activity contexts and this code-path is not properly using them.

The error happens inside of a method that has `TODO: how to handle failures?` on top of it, which I'm guessing is masking the failure on the specs.

If possible I'd add documentation to not use `execute_locally` outside of tests. This method caused issues for us before due to the lack of proper error handling.

jeffschoner:
### Summary

#234 broke local activities because the `Activity::Context` constructor callsite they rely on was not updated. Local activities were also not covered by any specs, which made it easy to miss this. This change passes the correct number of parameters to the constructor and adds a basic spec for local activity invocation.

Fixes #245 

### Testing

New spec: `bundle exec rspec spec/unit/lib/temporal/workflow/context_spec.rb:100`

### DIFF
--- a/lib/temporal/workflow/context.rb
+++ b/lib/temporal/workflow/context.rb
@@ -1,5 +1,6 @@
 require 'securerandom'
 
+require 'temporal/activity/context'
 require 'temporal/execution_options'
 require 'temporal/errors'
 require 'temporal/thread_local_context'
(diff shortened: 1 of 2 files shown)


antstorm (approved):
👍

DeRauk (approved):

PR #252, opened 2023-06-19
Fix task queue type to match enum

awesomenix:
Fix task queue type to match enum defined constant

When using `describe_task_queue` api, we run into an error

```
`describe_task_queue': uninitialized constant Temporalio::Api::Enums::V1::TaskQueueType::Workflow (NameError)
```


DeRauk (approved):

DeRauk:
Thank you!

PR #254, opened 2023-07-04
Allow setting credentials as lambda

laurynas:
We had a problem with multiple Puma workers and preload, because Temporal initializer which configures connection credentials ran before fork. Therefore after fork it we got an error:

```
grpc cannot be used before and after forking
```

Reconfiguring credentials after forking does not work because of the guard in grpc gem:
https://github.com/grpc/grpc/blob/e20e7a1198b2ff276a2f4f8514201b8ac1456830/src/ruby/ext/grpc/rb_grpc.c#L264

Therefore the most elegant solution was to configure credentials on demand, when connection is being built (after fork). 
I have added a lambda support to solve this issue. It works for us.

@egletam

### DIFF
--- a/lib/temporal/configuration.rb
+++ b/lib/temporal/configuration.rb
@@ -15,7 +15,8 @@ module Temporal
     Execution = Struct.new(:namespace, :task_queue, :timeouts, :headers, :search_attributes, keyword_init: true)
 
     attr_reader :timeouts, :error_handlers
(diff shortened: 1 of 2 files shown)


egletam (approved):

lorines:
@laurynas @egletam any updates on when this will be merged?

laurynas:
@lorines I don't know, it is up to the library maintainers.
cc @jeffschoner

gytisgreitai:
Hey @antstorm  anyone willing to take a look at this?

PR #256, opened 2023-07-11
Add missing arguments to Temporal.reset_workflow and add integration tests for Resets

jazev-stripe:
This PR adds two arguments present on the [`ResetWorkflowExecutionRequest` Protobuf message](https://api-docs.temporal.io/#temporal.api.workflowservice.v1.ResetWorkflowExecutionRequest) to the `Temporal.reset_workflow` API:
- `request_id` (an idempotency key)
- `reset_reapply_type` (used to optionally replay signals when resetting workflows)

This PR also adds integration tests for workflow resets, including tests for each of the above added arguments.

### Note: backwards compatibility

I attempted to make the API changes backwards compatible. However, I'm not sure if the `Temporal.reset_workflow` is used much in production, anyways, because I believe it has been broken ever since [Temporal Server v1.14.0](https://github.com/temporalio/temporal/releases/tag/v1.14.0) made the change that the `request_id` field must not be empty (in https://github.com/temporalio/temporal/pull/2141). Before my PR, temporal-ruby wasn't setting this field on the Protobuf object, causing it to use the zero value (empty string). This caused the Temporal Server to always respond with a gRPC "invalid argument" error.

I fixed this in this PR by falling back to a random UUID if the `request_id` is not specified. This is consistent with what the Go SDK does ([source](https://github.com/temporalio/sdk-go/blob/e1d76b7c798828302980d483f0981128c97a20c2/internal/internal_workflow_client.go#L952-L972)).

#### Reproduction of broken behavior

This can be reproduced by:

1. Checking out any tag on https://github.com/temporalio/docker-compose **after v1.14.0 (inclusive)**, and running `docker-compose up`
2. Creating the example namespace (`(cd examples && ./bin/register_namespace ruby-samples)`)
3. Running the example worker (`(cd examples && ./bin/worker)`
4. Running `(cd examples && ./bin/trigger QueryWorkflow)`. Note the workflow_id for the workflow this starts:
    ```sh
    $ (cd examples && ./bin/trigger QueryWorkflow)                                                                                                                                                                                                    
    I, [2023-07-11T16:22:25.593108 #603]  INFO -- temporal_client: Started workflow {"workflow_id":"bd442477-7348-428e-b40f-89793471a071","run_id":"14b8f6f5-6e6b-4f8d-ba96-d6c449fdbb34"}
    
    # Store the workflow_id:
    $ workflow_id="bd442477-7348-428e-b40f-89793471a071"
    ```

5. Running `(cd examples && ./bin/reset ruby-samples "$workflow_id")`:
    ```
    $ (cd examples && ./bin/reset ruby-samples "$workflow_id")
    Traceback (most recent call last):
    	...
    .../gems/grpc-1.56.0/src/ruby/lib/grpc/generic/active_call.rb:29:in `check_status': 3:RequestId is not set on request.. debug_error_string:{UNKNOWN:Error received from peer  {grpc_message:"RequestId is not set on request.", grpc_status:3, created_time:"2023-07-11T16:20:36.345632-07:00"}} (GRPC::InvalidArgument)
    ```

### DIFF
--- a/examples/spec/integration/reset_workflow_spec.rb
+++ b/examples/spec/integration/reset_workflow_spec.rb
@@ -1,0 +1,163 @@
+require 'workflows/hello_world_workflow'
+require 'workflows/query_workflow'
+require 'temporal/reset_reapply_type'
+
+describe 'Temporal.reset_workflow' do
+  it 'can reset a closed workflow to the beginning' do
+    workflow_id = SecureRandom.uuid
+    original_run_id = Temporal.start_workflow(
+      HelloWorldWorkflow,
+      'Test',
+      options: { workflow_id: workflow_id }
+    )
+
+    original_result = Temporal.await_workflow_result(
+      HelloWorldWorkflow,
+      workflow_id: workflow_id,
+      run_id: original_run_id
+    )
+    expect(original_result).to eq('Hello World, Test')
+
+    new_run_id = Temporal.reset_workflow(
+      Temporal.configuration.namespace,
+      workflow_id,
+      original_run_id,
+      strategy: Temporal::ResetStrategy::FIRST_WORKFLOW_TASK
+    )
+
+    new_result = Temporal.await_workflow_result(
+      HelloWorldWorkflow,
+      workflow_id: workflow_id,
+      run_id: new_run_id,
+    )
+    expect(new_result).to eq('Hello World, Test')
+  end
+
+  def reset_hello_world_workflow_10_times(workflow_id, original_run_id, request_id:)
+    10.times.map do
+      new_run_id = Temporal.reset_workflow(
+        Temporal.configuration.namespace,
+        workflow_id,
+        original_run_id,
+        strategy: Temporal::ResetStrategy::FIRST_WORKFLOW_TASK,
+        request_id: request_id
+      )
+
+      new_result = Temporal.await_workflow_result(
+        HelloWorldWorkflow,
+        workflow_id: workflow_id,
+        run_id: new_run_id,
+      )
+      expect(new_result).to eq('Hello World, Test')
+
+      new_run_id
+    end
+  end
+
+  it 'can repeatedly reset the same closed workflow to the beginning' do
+    workflow_id = SecureRandom.uuid
+    original_run_id = Temporal.start_workflow(
+      HelloWorldWorkflow,
+      'Test',
+      options: { workflow_id: workflow_id }
+    )
+
+    original_result = Temporal.await_workflow_result(
+      HelloWorldWorkflow,
+      workflow_id: workflow_id,
+      run_id: original_run_id,
+    )
+    expect(original_result).to eq('Hello World, Test')
+
+    new_run_ids = reset_hello_world_workflow_10_times(
+      workflow_id,
+      original_run_id,
+      # This causes the request_id to be generated with a random value:
+      request_id: nil
+    )
+
+    # Each Reset request should have resulted in a unique workflow execution
+    expect(new_run_ids.uniq.size).to eq(new_run_ids.size)
+  end
+
+  it 'can deduplicate reset requests' do
+    workflow_id = SecureRandom.uuid
+    original_run_id = Temporal.start_workflow(
+      HelloWorldWorkflow,
+      'Test',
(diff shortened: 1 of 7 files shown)


examples/spec/integration/reset_workflow_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+    new_result = Temporal.await_workflow_result(
+      HelloWorldWorkflow,
+      workflow_id: workflow_id,
+      run_id: new_run_id,
+    )
+    expect(new_result).to eq('Hello World, Test')
+  end
+
+  def reset_hello_world_workflow_10_times(workflow_id, original_run_id, request_id:)
+    10.times.map do
```
jeffschoner, line 37: Is 10 really necessary? Two should catch any idempotency issues, right? Beyond that seems more like testing the Temporal server rather than the SDK.

lib/temporal/connection/grpc.rb
```diff
@@ -409,16 +414,33 @@ def signal_with_start_workflow_execution(
         client.signal_with_start_workflow_execution(request)
       end
 
-      def reset_workflow_execution(namespace:, workflow_id:, run_id:, reason:, workflow_task_event_id:)
+      def reset_workflow_execution(namespace:, workflow_id:, run_id:, reason:, workflow_task_event_id:, request_id: nil, reset_reapply_type: nil)
+        if request_id.nil?
```
jeffschoner, line 418: I'd put this in the layer above since the GRPC/connection layer is mostly used for mapping payloads

lib/temporal/client.rb
```diff
@@ -281,9 +281,13 @@ def await_workflow_result(workflow, workflow_id:, run_id: nil, timeout: nil, nam
     # @param workflow_task_id [Integer, nil] A specific event ID to reset to. The event has to
     #   be of a type WorkflowTaskCompleted, WorkflowTaskFailed or WorkflowTaskTimedOut
     # @param reason [String] a reset reason to be recorded in workflow's history for reference
+    # @param request_id [String, nil] an idempotency key for the Reset request or `nil` to use
+    #   an auto-generated, unique value
+    # @param reset_reapply_type [Symbol, nil] one of the Temporal::ResetReapplyType values or `nil`
```
jeffschoner, line 286: I assume this maps to the unspecified enum value? Does Temporal make any claim about how that will behave? You might instead pick `None` as the default and require a valid enum value here.

lib/temporal/connection/grpc.rb
jazev-stripe: I think when I was deciding where to put this behavior, I wasn't sure if the `Temporal::Connection::GRPC.*` methods were considered a part of the public API, so I didn't want to add new required arguments to it.

However, with the fact that it's broken on Temporal Server versions since v1.14, I feel like adding this as a required arguments may actually be a good thing, because it reflects that new server versions also require it. It's also possible that this method isn't considered a part of the public API of the gem, so maybe it doesn't matter.

examples/spec/integration/reset_workflow_spec.rb
jazev-stripe: Fair enough, 10 was just an arbitrary number; I can replace it with 2.

lib/temporal/client.rb
jazev-stripe: Exactly, yeah it maps to unspecified.

There aren't any claims about it in the gRPC API contract, and to be honest there are practically no docs on the semantics of Reset (that I could find).

Looking at the server implementation it looks like this maps to `:signal` as the default ([source](https://github.com/yiminc/temporal/blob/30a89202af4f508def300c4ae6232f1045be90a1/common/enums/defaults.go#L67)), so I'm inclined to make that the default.

lib/temporal/connection/grpc.rb
jeffschoner-stripe: I don't believe it's intended to be used directly. Since the argument is required by Temporal server, making it required at this level seems to make sense too.

jeffschoner: I don't believe it's intended to be used directly. Since the argument is required by Temporal server, making it required at this level seems to make sense too.

PR #261, opened 2023-08-26
Signals first ordering

### LINKED ISSUE #258: Signals cannot be processed in final workflow task
### Description
If it's unacceptable for signals sent to workflow to be lost, all outstanding signals must be processed before a workflow returns. If signals arrive while this final workflow task is being processed, the server will fail the workflow task, add the signal to the history, and schedule a new workflow task. This behavior guarantees that workflow code is able to run on the signal history event. Today, however, that signal cannot be processed because its dispatch handlers will only be invoked after the workflow has returned.

### Repro
Consider this example workflow code:
```
class ReproWorkflow < Temporal::Workflow
  def execute(sleep_for)
    signals = []
    workflow.on_signal do |signal|
      signals << signal
    end

    workflow.sleep(sleep_for)
    signals.count
  end
end
```

invoked in this test,
```
  it 'signals repro workflow' do
    workflow_id = SecureRandom.uuid
    run_id = Temporal.start_workflow(
      ReproWorkflow,
      1, # seconds
      options: {
        workflow_id: workflow_id,
        signal_name: 'a_signal',
        signal_input: 'foo',
        timeouts: { execution: 10 },
      }
    )

    sent = 1
    loop do
      begin
        Temporal.signal_workflow(ReproWorkflow, 'a_signal', workflow_id, run_id, 'foo')
      rescue
        # Keep going until there's an error such as the workflow finishing
        break
      end
      sent += 1
      sleep 0.01
    end

    result = Temporal.await_workflow_result(
      SignalWithStartWorkflow,
      workflow_id: workflow_id,
      run_id: run_id,
    )

    expect(result).to eq(sent)
  end
```

Although the exact results can vary because there is a race condition here, almost always this test fails like:
```
Failures:

  1) signal with start signals repro workflow
     Failure/Error: expect(result).to eq(sent)

       expected: 78
            got: 73

       (compared using ==)
     # ./spec/integration/signal_with_start_spec.rb:108:in `block (2 levels) in <top (required)>'
```

Looking at the workflow history ([JSON](https://github.com/coinbase/temporal-ruby/files/12262696/repro-history.txt) collected during a test run), you'll see a workflow task failure of `UnhandledCommand`. This is an expected failure which ensures all signals can be processed mentioned above. When the code is evaluated on replay, it exits before processing the last batch of signals following the failed task. If a debugger is placed in the code, it's clear that the workflow function returns before the signal handlers fire.

In this example, the signal handlers only modify state that determines what will be returned. However, if these run activities, timers, or other workflow code, other errors complaining about commands being performed after a workflow has returned can occur, causing the workflow task to perpetually fail and put it into a "stuck" state where it cannot complete.

### Root cause and solutions
This seems to be happening because signal dispatch handlers are invoked in history order. Other Temporal SDKs evaluate all signal history entries within a history window first. temporal-ruby does this for markers today, but also needs to do it for signals. I confirmed this is the expected behavior with Chad Retz from Temporal.

Making this change at its core is fairly simple. However, it will affect determinism of all existing workflows that use signals. In more recent versions of Temporal server, there is now an SDK metadata field available on the workflow task completion request ([source](https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L324) [source](https://github.com/temporalio/api/blob/master/temporal/api/sdk/v1/task_complete_metadata.proto#L34)). This field can be set by the SDK to indicate which features were used in evaluating a workflow task, and read it from workflow task completed history entries to ensure deterministic behavior changes are safe. Moreover, because older versions of the SDK will not be able to understand these SDK flags and will have different signal determinism, a compatibility configuration option will need to be added so that the new version can be safely rolled out "dark" before being switched on.

jeffschoner:
### Summary 
Introduces signals first ordering within a workflow task window. Other Temporal SDKs handle markers and signals before other event types when replaying workflow history. This ensures that all signals can be processed before a workflow completes, fails, or continues as new.

Fixes #258 

### Important Note
If you have multiple workers in a fleet and use rolling deployments, adopting this change safely requires additional steps to avoid workflow history corruption or "stuck" workflows.  See `CHANGELOG.md` for details on how to use the `legacy_signals` configuration flag to prevent these problems during rollout of this version.

### Details
With this change, temporal-ruby now handles markers and signals first, as opposed to just markers first. There are a number of parts to this change:
- **Refactored the code that orders events within a history window.** Previously, it separated out markers, then evaluated those before other events. It now sorts events based a custom sort function that considers markers, signals, and other event types.
- **Adds SDK metadata support for workflow tasks**, and a new language flag for this ordering behavior. This ensures that workflow code run on the new or old ordering, continues to execute deterministically. This can be used for future determinism breaking changes.
- **Adds lazy loading of capabilities** via `GetSystemInfo`. The SDK metadata feature this new ordering relies on is only available on Temporal server 1.20.0 or newer. This ensures the feature will not be activated when connected to an older server version where this change cannot be safely used.
- **Adds a `legacy_signals` configuration option** to preserve old ordering behavior. This serves a few important purposes:
  - Delaying adoption until a new server version has been completely rolled out
  - Non-determinism issues when deploying this code for the first time
- Some files have styling changes from the use of `rubyfmt` 

### Testing
- Many new and updated specs for worker, task processor, GRPC, and state manger
- A new integration spec for signals that is fixed by this change


lib/temporal/workflow/state_manager.rb
```diff
@@ -83,43 +89,98 @@ def apply(history_window)
...
-        history_window.markers.each do |event|
+        order_events(history_window.events).each do |event|
           apply_event(event)
         end
+      end
 
-        history_window.events.each do |event|
-          apply_event(event)
+      def self.event_order(event, signals_first)
+        if event.type == 'MARKER_RECORDED'
```
mjameswh, line 100: Note that this first category is only for _patches_. Other type of markers (eg. local activities and side effects) need to be sorted in the "everything else" category.

lib/temporal/workflow/state_manager.rb
```diff
@@ -83,43 +89,98 @@ def apply(history_window)
...
+          # markers always come first
+          0
+        elsif event.type == 'WORKFLOW_EXECUTION_STARTED'
+          # This always comes next if present
+          1
+        elsif signals_first && signal_event?(event)
+          # signals come next if we are in signals first mode
+          2
+        else
```
mjameswh, line 109: You also need to sort _queries_ in a distinct category, after everything else.

lib/temporal/workflow/state_manager.rb
jeffschoner: The current behavior (with and without my change) is that all markers are sorted first. It sounds like only the release markers (similar to but not the same as patches in the Rust core SDKs) should come first. I believe the only other supported marker types are for side effects and local activities. What is the harm of those coming first? I believe they should both be ok since their callbacks aren't invoked on replay (only when playing for the first time) and because they shouldn't (at least in theory) otherwise modify state in workflow functions.

lib/temporal/workflow/state_manager.rb
jeffschoner: I don't believe there is any code that handles query history events specifically. Actually, I don't see any history events that have to do with queries that would be handled here. All query handling code lives at the task processor level, and doesn't involve anything in the state manager as far as I know.

@dwillett What do you think?

lib/temporal/workflow/state_manager.rb
mjameswh: > What is the harm of those coming first?

Well... That really depends how strict you are about matching history commands on replay. In Core-based SDKs, we expect all commands to replay in the very same order (not sure about Go and Java).

Now, I just looked at how side effect markers are handled, and they turn out getting stored in a distinct queue, from which you pull every time workflow calls the `side_effect` method. That mostly achieve the same as if they had been left in the "everything else" bucket, except that ordering of side effects vs other type of commands is lost. So that's probably ok, given that side_effects handlers are called synchronously, so ordering of other stuff can't be messed up.

DeRauk (approved):

PR #268, opened 2023-10-15
Save signals on first workflow task

### LINKED ISSUE #267: Cannot handle signals on first workflow task
Signals received before the first workflow task are effectively not available to code running in the first workflow task. Notably, this happens when using start-with-signal.

This happens because since #261 events within the first workflow task are handled in the following order:
1. Markers for versions and side effects
2. Start workflow execution. This is always the first history entry in every workflow run. This causes workflow code to start executing.
3. Signals

Although 'start workflow execution' starts workflow code which can register a signal handler, no signals are invoked against those handlers until after the workflow code deschedules, such as when encountering a sleep or waiting for an activity to complete.

For example, consider the following code:
```ruby
class ReproWorkflow < Temporal::Workflow
  def execute
    received = nil
    workflow.on_signal do |signal, input|
      received = input
    end

    if received.present?
      'got signal'
    else
      'no signal'
    end
end

# Start this with a signal
Temporal.start_workflow(ReproWorkflow, {signal_name: 'foo', signal_input: 'bar'})
```
The workflow started here will always return `'no signal'` because `received` will not be updated until after the workflow function has already returned.

This problem does not exist outside of the first workflow task because the signal handler can simply be registered earlier in the workflow. It's not possible to pre-register a signal handler before execution.

Workarounds include:
- Adding code that deschedules the workflow between signal handler registration and reading any of its effects, such as sleeping for 1 second before checking the value of `received` above
- Accepting that signals will be delivered later in a workflow's execution than expected

Neither of these are great solutions, but they do limit the severity of this bug.

jeffschoner:
This fixes #267. See that issue for more details on the bug.

This change makes signals available to code running in the first workflow task by saving any signals received before the first workflow task, then calling any signal handler registered during the first workflow task with these saved values. Because this change would otherwise affect the determinism of all workflows that start workflows with a signal, it adds a new SDK flag to preserve the previous behavior for existing workflow histories.

This is achieved by registering a special signal handler that saves signals during the first workflow task. Any signal handlers registered during the first workflow task will be immediately called with any of these saved signals. After code passes the first workflow task, the handler is removed and behavior returns to normal.


jeffschoner:
This still needs an optional configuration setting for disabling this behavior. It's not functionally necessary, but it provides more control over rollout.

DeRauk (approved):

PR #269, opened 2023-10-16
History size and suggest continue as new

jeffschoner:
Starting in Temporal server 1.20.0 (from https://github.com/temporalio/temporal/pull/3055), workflow task started history events contain attribute fields indicating the current history size (in bytes and in event count) and a bit indicating if the workflow should continue as new. This change surfaces this information on the workflow context.

This change includes a new example `ContinueAsNewWorkflow` showing how to use this.


DeRauk (approved):

DeRauk:
Thanks @jeffschoner! This is a great new feature.

PR #270, opened 2023-10-17
Add option for gRPC client connection retries

hughevans:
Add a config option for a basic gRPC connection retry policy:

``` ruby
config.connection_options = {
  retry_connection: true
}
```

That uses a basic policy:

```ruby
{
  retryableStatusCodes: ["UNAVAILABLE"],
  maxAttempts: 3,
  initialBackoff: "0.1s",
  backoffMultiplier: 2.0,
  maxBackoff: "0.3s"
}
```

There’s also a `retry_policy` option to set a custom policy if you want more control:

``` ruby
config.connection_options = {
  retry_policy: {
    retryableStatusCodes: ["UNAVAILABLE", "INTERNAL"],
    maxAttempts: 5,
    initialBackoff: "0.1s",
    backoffMultiplier: 2.5,
    maxBackoff: "0.7s"
  }
}
```

### DIFF
--- a/lib/temporal/connection/grpc.rb
+++ b/lib/temporal/connection/grpc.rb
@@ -804,8 +804,16 @@ module Temporal
           channel_args["grpc.keepalive_time_ms"] = options[:keepalive_time_ms]
         end
 
-        if options[:retry_connection]
+        if options[:retry_connection] || options[:retry_policy]
           channel_args["grpc.enable_retries"] = 1
+
+          retry_policy = options[:retry_policy] || {
+            retryableStatusCodes: ["UNAVAILABLE"],
+            maxAttempts: 3,
+            initialBackoff: "0.1s",
+            backoffMultiplier: 2.0,
+            maxBackoff: "0.3s"
+          }
 
           channel_args["grpc.service_config"] = ::JSON.generate(
             methodConfig: [
@@ -815,13 +823,7 @@ module Temporal
                     service: "temporal.api.workflowservice.v1.WorkflowService",
                   }
                 ],
-                retryPolicy: {
-                  retryableStatusCodes: ["UNAVAILABLE"],
-                  maxAttempts: 3,
-                  initialBackoff: "0.1s",
-                  backoffMultiplier: 2.0,
-                  maxBackoff: "0.3s"
-                }
+                retryPolicy: retry_policy
               }
             ]
           )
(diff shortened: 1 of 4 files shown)


Mikh333 (approved):

lib/temporal/connection/grpc.rb
```diff
@@ -633,11 +634,43 @@ def get_system_info
...
+                  {
+                    service: "temporal.api.workflowservice.v1.WorkflowService",
+                  }
+                ],
+                retryPolicy: {
+                  retryableStatusCodes: ["UNAVAILABLE"],
+                  maxAttempts: 3,
+                  initialBackoff: "0.1s",
+                  backoffMultiplier: 2.0,
+                  maxBackoff: "0.3s"
```
DeRauk, line 661: @hughevans thanks for the PR! It looks good overall.

For maxAttempts, initialBackoff, backoffMultiplier, and maxBackoff - I think these are sane defaults, but if we're going to support retries I'd prefer that users be able to pass in their own values to override these if they want. Could you make them configurable?

hughevans: @DeRauk Thank you for checking this out. I’ve gone ahead and added a `retry_policy` option there too.

Also I looked at the README, but was a bit unsure where exactly documentation for this should go logically. Is that something you could take on after merging possibly? 🙇

spec/unit/lib/temporal/grpc_spec.rb
```diff
@@ -830,4 +830,101 @@ class TestDeserializer
...
+              ]
+            }.to_json
+          }
+        )
+        subject.send(:client)
+      end
+    end
+
+    context "when passing a custom retry policy" do
+      subject { Temporal::Connection::GRPC.new(nil, nil, identity, :this_channel_is_insecure, retry_policy:) }
```
DeRauk, line 892: @hughevans typo with `retry_policy:)`?

lib/temporal/connection/grpc.rb
DeRauk: Thanks, and sure.

spec/unit/lib/temporal/grpc_spec.rb
hughevans: Ah, sorry that only works in newer versions of Ruby. I have pushed a fix now thanks.

DeRauk (approved):

PR #272, opened 2023-10-17
Expose count_workflow_executions on the temporal client

harsh-stripe:
# Summary
This change exposes a `count_workflow_executions` method on the Temporal client so that users can invoke it using code such as:
```
client = Temporal::Client.new(config)
result = client.count_workflow_executions(namespace: 'ruby-samples', query: 'WorkflowType="TestingWorkflow"')
```

# Test Plan
Unit test:
```
bundle exec rspec spec/unit/lib/temporal/client_spec.rb:1106
```


lib/temporal/workflow/count_workflows_aggregation.rb
```diff
@@ -0,0 +1,11 @@
+module Temporal
+  class Workflow
+    class CountWorkflowAggregation
```
jazev-stripe, line 3: ooc, why "Aggregation" as the suffix instead of a more neutral "Result" or "Response"? Does aggregation mean it is aggregating all of the fields in the `CountWorkflowExecutionsResponse`, or is it a reference to the `CountWorkflowExecutionsResponse.AggregationGroup` type?

spec/unit/lib/temporal/client_spec.rb
```diff
@@ -1102,4 +1102,28 @@ class NamespacedWorkflow < Temporal::Workflow
...
+    end
+
+    before do
+      allow(connection)
+        .to receive(:count_workflow_executions)
+        .and_return(response)
+    end
+
+    it 'returns the count' do
+      resp = subject.count_workflow_executions(namespace, '')
```
jazev-stripe, line 1120: nit: having a non-empty query makes the test slightly more realistic (at least I don't think an empty query works on the real server)

lib/temporal/client.rb
```diff
@@ -425,6 +426,11 @@ def query_workflow_executions(namespace, query, filter: {}, next_page_token: nil
       Temporal::Workflow::Executions.new(connection: connection, status: :all, request_options: { namespace: namespace, query: query, next_page_token: next_page_token, max_page_size: max_page_size }.merge(filter))
     end
 
+    def count_workflow_executions(namespace, query)
```
jazev-stripe, line 429: it would be nice to have a short doc-comment with the types of the parameters (even if it isn't enforced without Sorbet, and in this case the types are fairly trivial). I know a lot of the other methods don't have it, but I'm putting my type evangelist hat on 🤠.

jazev-stripe (changes requested):
I only had a few small ideas; let me know if you disagree

spec/unit/lib/temporal/client_spec.rb
harsh-stripe: Believe it or not, an empty query _does_ work on the real server (see slack), but I've gone ahead and modified the test to be more realistic (there's very few scenarios during which you'd do an empty query).

lib/temporal/client.rb
```diff
@@ -425,6 +426,17 @@ def query_workflow_executions(namespace, query, filter: {}, next_page_token: nil
       Temporal::Workflow::Executions.new(connection: connection, status: :all, request_options: { namespace: namespace, query: query, next_page_token: next_page_token, max_page_size: max_page_size }.merge(filter))
     end
 
+    # Count the number of workflows matching the provided query
+    # 
+    # @param namespace [String]
+    # @param query [String]
+    #
+    # @return [Temporal::Workflow::CountWorkflowAggregation] an integer count of workflows matching the query
```
jazev-stripe, line 434: very small typo:
```suggestion
    # @return [Temporal::Workflow::CountWorkflowsResult] an integer count of workflows matching the query
```

jazev-stripe (approved):
The only bit of feedback was the typo in the doc-comment. Other than that, it looks great

lib/temporal/client.rb
```diff
@@ -425,6 +426,17 @@ def query_workflow_executions(namespace, query, filter: {}, next_page_token: nil
...
 
+    # Count the number of workflows matching the provided query
+    # 
+    # @param namespace [String]
+    # @param query [String]
+    #
+    # @return [Temporal::Workflow::CountWorkflowsResult] an integer count of workflows matching the query
+    def count_workflow_executions(namespace, query: nil)
+      response = connection.count_workflow_executions(namespace: namespace, query: query)
+      Workflow::CountWorkflowsResult.new(count: response.count)
```
DeRauk, line 437: @harsh-stripe Why create a new response object to return here? It'd be simpler to just return the count from this method

harsh-stripe: Made a commit to return the count instead of this wrapped object. I had initially created it thinking it might be useful for a future contributor if they wanted to include aggregation groups based on https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L795-L798 in the future, but it isn't required now since it doesn't quite work yet in Temporal.

harsh-stripe:
@DeRauk - Addressed your feedback to return the `count` value directly instead of as a wrapped object.

DeRauk (approved):

harsh-stripe:
@DeRauk - it looks like the [GitHub actions](https://github.com/coinbase/temporal-ruby/blob/master/.github/workflows/tests.yml) for this project sets up a temporal-server instance which does not use advanced visibility.

Ultimately, I think it might just be worth removing the example spec for `count_workflow_executions` since the spec relies on a visibility backend store, and it is also reliant on timing since the data push to the visibility store happens async. The spec would've been nice to have, but I don't believe it's critical, so I've removed it.

DeRauk (approved):

DeRauk:
Thank you!

PR #277, opened 2023-12-14
Add Schedule api support

dhruv-stripe:
## Summary

This PR adds support for interacting with the [Temporal Schedules Api](https://temporal.io/blog/how-we-build-it-building-scheduled-workflows-in-temporal). We're using this patch internally to synchronize schedules with Temporal when new code is deployed.

The approach I took was to define a set of `Temporal::Schedule` classes that roughly map to the proto models defined in the [temporal api](https://github.com/temporalio/api/blob/master/temporal/api/schedule/v1/message.proto).

This is a large PR given the surface area of the api, but I've tried to break it by commit to make it a bit easier to review.

## Test Plan

Ran `rspec spec` and `cd examples && rspec spec` without error

### DIFF
--- a/spec/unit/lib/temporal/connection/converter/payload/proto_json_spec.rb
+++ b/spec/unit/lib/temporal/connection/converter/payload/proto_json_spec.rb
@@ -29,14 +29,4 @@ describe Temporal::Connection::Converter::Payload::ProtoJSON do
 
     expect(subject.to_payload(input)).to be nil
   end
-
-  # DO NOT MERGE THIS UPSTREAM TO COINBASE
-  it 'exemptions' do
-    data = '{"name":"foo"}'
-    Temporal::Connection::Converter::Payload::ProtoJSON::SPECIAL_STRIPE_WORKFLOW_PAYLOAD_TYPES.each do |message_type|
-      fake_payload = Struct.new(:metadata, :data).new({ 'messageType' => message_type }, data)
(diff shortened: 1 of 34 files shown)


spec/unit/lib/temporal/connection/converter/payload/proto_json_spec.rb
```diff
@@ -29,4 +29,14 @@
...
+
+  # DO NOT MERGE THIS UPSTREAM TO COINBASE
+  it 'exemptions' do
+    data = '{"name":"foo"}'
+    Temporal::Connection::Converter::Payload::ProtoJSON::SPECIAL_STRIPE_WORKFLOW_PAYLOAD_TYPES.each do |message_type|
+      fake_payload = Struct.new(:metadata, :data).new({ 'messageType' => message_type }, data)
+      data_out = subject.from_payload(fake_payload)
+      expect(data_out).to eq({ 'name' => 'foo' })
+    end
+  end
```
jeffschoner, line 41: @dhruv-stripe These lines from Stripe's fork leaked through in your merge and are causing CI to fail

jeffschoner, line 41:
```suggestion
```

DeRauk (approved):

DeRauk:
Thank you!

PR #286, opened 2024-02-15
[WIP] Allow inheritance of retry and timeout options

hughevans:
Presently you only have the option of using the default options or per-class configuration of timeout and retry. This PR changes it so subclassed executables inherit these settings from their parent Ruby class.


hughevans:
Looks like I have more work to do.

Issue #295, opened 2024-03-29
:use_error_serialization_v2 config option isn't respected when using a custom Temporal::Configuration


davehughes:
This was causing me some confusion and I tracked it down to [this line](https://github.com/coinbase/temporal-ruby/blob/c4fb094638c85289c16b4611c3432f759cfb13a8/lib/temporal/connection/grpc.rb#L314), which uses the global singleton `Temporal.configuration` rather than the specific configuration that was used to create the connection.

I'm working on a PR to fix that:
1. adds `use_error_serialization_v2` to the set of options respected by Temporal::Connection::GRPC.
2. plumbs this value through from [Temporal::Configuration#for_connection](https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/configuration.rb#L118) which is used when actually creating the connection object in [Temporal::Client#connection](https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/client.rb#L550).

PR #296, opened 2024-03-29
Propagate :use_error_serialization_v2 option to GRPC client

davehughes:
Attempt to address https://github.com/coinbase/temporal-ruby/issues/295. 

Alter `Temporal::Connection::GRPC` to respect a `:use_error_serialization_v2` option, falling back to the global singleton value if not found, and update `Temporal::Configuration#for_connection` to pass its setting through when creating connections.

It looks like options can already be plumbed through by editing a Temporal::Configuration instance's `connection_options`, but this case seems slightly different because `use_error_serialization_v2` is an explicit setting on the configuration.

### DIFF
--- a/lib/temporal/configuration.rb
+++ b/lib/temporal/configuration.rb
@@ -122,7 +122,7 @@ module Temporal
         port: port,
         credentials: credentials,
         identity: identity || default_identity,
-        connection_options: connection_options
+        connection_options: connection_options.merge(use_error_serialization_v2: @use_error_serialization_v2)
       ).freeze
     end
 
--- a/lib/temporal/connection/grpc.rb
+++ b/lib/temporal/connection/grpc.rb
@@ -311,7 +311,7 @@ module Temporal
       end
 
       def respond_activity_task_failed(namespace:, task_token:, exception:)
-        serialize_whole_error = Temporal.configuration.use_error_serialization_v2
+        serialize_whole_error = options.fetch(:use_error_serialization_v2, Temporal.configuration.use_error_serialization_v2)
         request = Temporalio::Api::WorkflowService::V1::RespondActivityTaskFailedRequest.new(
           namespace: namespace,
           identity: identity,


DeRauk:
@davehughes Thank you for the PR! Could you please rebase onto master to get the test fix for CI?

njaremko:
Bump?

DeRauk (approved):

davehughes:
@DeRauk thanks!

jeffschoner:
@davehughes @DeRauk I'm not sure this is working as expected. I believe the value set in the configuration instance will always be used in `grpc.rb`. The `:use_error_serialization_v2` option cannot be unset such that it will fall back to the global. Setting it to `false` or `nil` does not seem to change this behavior.

E.g.,
```
irb(main):001:0> options = {}
=> {}
irb(main):002:0> options.fetch(:use_error_serialization_v2, true)
=> true
irb(main):003:0> options = {}.merge(use_error_serialization_v2: nil)
=> {:use_error_serialization_v2=>nil}
irb(main):004:0> options.fetch(:use_error_serialization_v2, true)
=> nil
irb(main):005:0> options = {}.merge(use_error_serialization_v2: false)
=> {:use_error_serialization_v2=>false}
irb(main):006:0> options.fetch(:use_error_serialization_v2, true)
=> false
```

Reading this from the options that are passed in from the configuration is still the right way to go, rather than relying on global configuration being read via `Temporal.configuration`.

This change, however, does make a breaking change for anyone who has opted into error serialization v2 via global config. 

What do you think about changing this from
```ruby
options.fetch(:use_error_serialization_v2, Temporal.configuration.error_serialization_v2)
```
to
```ruby
options[:use_error_serialization] || Temporal.configuration.error_serialization_v2
```
?

davehughes:
Hmm...I see what you mean and how this is an accidentally breaking change for a set of scenarios.

My mental model of configuration usage here is that folks should either choose to configure via the global singleton or through local configuration objects depending on their needs. I'm not sure how strongly the project wants to hold to the semantics that the global config is a fallback vs. a distinct thing, though clearly there's some of the former going on. 

I don't think the proposed change works as written, since if `options[:use_error_serialization] == false`, the code will always fall back to the global setting. But allowing `nil` values and falling back in the presence of a `nil` should work:
```
use_v2 = options[:use_error_serialization_v2]
use_v2 = Temporal.configuration.error_serialization_v2 if use_v2.nil?
```
In conjunction with using a `nil` default in the configuration constructor [here](https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/configuration.rb#L80) should allow the global fallback to apply as you expect.

All that said, I'm really just a nomad passing through this project. I've left the job where I noticed the issue that led to this PR, and fixing this is not really a priority for me. (Although I do apologize for the inadvertent breakage, and would be happy to shepherd through a PR with the approach above if asked.)

PR #300, opened 2024-05-28
Replay testing

jeffschoner:
### Summary

Adds a `ReplayTester` like the replayer abstractions in official SDKs. This allows workflow histories to be replayed against code for testing purposes to detect non-determinism before a change goes to production.

This consists of a set of smaller changes. These can be reviewed commit by commit:
- Specializes workflow event targets by making a mismatch between workflow completion, failure, or continue as new non-deterministic, whereas previously these were all considered equivalent. Moreover, this now catches non-determinism cases where a workflow prematurely ends on replay.
- Adds methods for downloading and working with histories in JSON and protobuf binary format
- The `ReplayTester` abstraction itself
- An example replay test for `SignalWithStartWorkflow` and a script to update its history that demonstrates collecting histories in JSON and protobuf binary
- Inclusion of fiber backtraces on failure so that the error message on replay failure is more expected and useful
- An option to log while replaying workflows. This is used by the replay tester to emit logs while running to provide better debuggability of tests, but it can be used in other circumstances as well

### Testing

- New unit specs for the replay tester
- Existing specs have been cleaned up, mostly around workflow event target specialization
- An example replay test


lib/temporal/testing/replay_tester.rb
```diff
@@ -0,0 +1,136 @@
...
+      # - "Unexpected event UNSPECIFIED". This can occur if the history was downloaded with Java-style enum values, which
+      #   can happen in certain versions of certain SDKs. Use the correct_event_types function below to convert these.
+      #   See that function's comment for more details.
+      def replay_history_json_file(workflow_class, path)
+        json = File.read(path)
+        replay_history_json(workflow_class, json)
+      end
+
+      # Runs a replay test on a JSON string directly. See comment on replay_history_json_file more details.
+      def replay_history_json(workflow_class, json)
```
drewhoskins-temporal, line 40: Feels a bit more composable to have just one typesafe `replay_history` method which takes a `History` and then have some helper methods to convert from proto/json/files to `History`
Was perusing the [python](https://github.com/temporalio/sdk-python/blob/58d695106a339e9e0be6fc02c78a4dcf57ce6eca/temporalio/worker/_replayer.py#L96) version, might be worth looking for insights there as well.

lib/temporal/testing/replay_tester.rb
```diff
@@ -0,0 +1,136 @@
...
+        # than being pulled via a workflow task, no query support, no metrics, and other
+        # simplifications. Fake metadata needs to be provided.
+        metadata = Temporal::Metadata::WorkflowTask.new(
+          namespace: 'replay-test',
+          id: 1,
+          task_token: '',
+          attempt: 1,
+          workflow_run_id: 'run_id',
+          workflow_id: 'workflow_id',
+          workflow_name: history.find_event_by_id(1).attributes.workflow_type.name
```
drewhoskins-temporal, line 82: nit: Could probably use a comment or assertion that this is the event you expect.

lib/temporal/testing/replay_tester.rb
```diff
@@ -0,0 +1,136 @@
...
+
+      # Protobuf does not define a consistent format for enums in JSON. In some Temporal SDKs, the event
+      # type enum takes the form 'WorkflowTaskScheduled' but here it needs to be EVENT_TYPE_WORKFLOW_TASK_SCHEDULED.
+      # This code parses the JSON, reformats these fields, then regenerates the JSON. This should not be necessary
+      # for histories downloaded from recent versions of the Temporal CLI, Temporal UI, or using methods
+      # on Temporal::Client.
+      #
+      # If the pretty_print optional parameter is set to true, it outputs in a more human
+      # readable form on output.
+      def self.correct_event_types(text, pretty_print: true)
```
drewhoskins-temporal, line 120: Is this because Stripe is on an old version of the Java SDK?  Don't see this function used anywhere.  Should it be tested?  Or maybe just done on the stripe fork?

lib/temporal/testing/replay_tester.rb
```diff
@@ -0,0 +1,136 @@
...
+  module Testing
+    class ReplayError < StandardError
+    end
+
+    class ReplayTester
+      def initialize(config: Temporal.configuration)
+        # Duplicate the configuration so that this doesn't interfere with other tests in the
+        # same process that are not replay tests
+        @config = config.dup.tap do |c|
+          # Otherwise, replay tests will produce no logs. This helps with test debugging.
```
drewhoskins-temporal, line 19: Interesting idea.  I like it, and I couldn't find evidence that the temporal SDKs do this.
I think people would want this in tests, but perhaps not if doing safe-rollout deploys (might be scary to see redundant logs during production rollouts).  I wonder if this should be configurable.

lib/temporal/workflow/executor.rb
```diff
@@ -33,7 +33,7 @@ def initialize(workflow_class, history, task_metadata, config, track_stack_trace
 
       def run
         dispatcher.register_handler(
-          History::EventTarget.workflow,
+          History::EventTarget.start_workflow,
```
drewhoskins-temporal, line 36: Curious what's going on here.  I ask because it's pretty central code.

lib/temporal/testing/replay_tester.rb
Sushisource: Could be a nice option to have during testing, certainly, but yeah I'd make it configurable defaulting to off.

lib/temporal/testing/replay_tester.rb
jeffschoner: I believe this originally started with older versions of tctl. Unfortunately, we standardized our internal tooling on this camel case enum style rather than the standard screaming snake case version. It probably does make more sense to leave it out of here, as any new histories collected using current Temporal UI/CLI or these new APIs won't need to rely on this. I agree if it's kept, it should be tested.

lib/temporal/workflow/executor.rb
jeffschoner: The non-determinism checking that happens in `discard_command` only considers the "target" of a command/event and only includes it and the event ID in its error messages. Prior to this change, any operation affecting a workflow had the same target and therefore the same target in any error messages. I've specialized these targets to individual ones for starting, completing, failing and continuing as new. This comes with two major benefits.

First, the non-determinism checking now applies between these different cases. This means a history that succeeded but on replay tries to fail or continue as new will now be considered non-deterministic while previously this would have been acceptable

Second, the error message will now differentiate between these cases rather than a generic "workflow" error that may be non-obvious to the user. E.g., this error message:
> Unexpected command. The replaying code is issuing: (23, workflow) but the history of previous executions recorded: (23, activity)

becomes something like,

> Unexpected command. The replaying code is issuing: (23, fail_workflow) but the history of previous executions recorded: (23, activity)

On top of that, we now call `discard_command` in the handlers for these workflow ending events, which means we do non-determinism checking on workflow end. This is important in catching cases where a workflow history ends but on replay the code keeps running.

This .particular change is simply a consequence of that specialization of event targets. I could have left it in place, but felt it would be more confusing to see a `workflow` target only for starting workflows, while the other cases all have their own targets.

jeffschoner:
I still plan to make some of the suggested improvements, but marking this as "ready to review" to make clear it's in a reviewable state.

lib/temporal/testing/replay_tester.rb
jeffschoner: It's not as close as it could be because temporal-ruby already has its own representation of workflow histories that I didn't want to refactor, but I was able to bring its more composable API shape over here

jeffschoner:
@drewhoskins-temporal @Sushisource Thanks for the feedback. I've incorporated your suggestions

@DeRauk Ready for your review too

DeRauk (approved):

DeRauk:
This is really great, thank you @jeffschoner!

PR #304, opened 2024-07-05
Add UpdatableTimer example

taonic:
Demonstrates how to use signals to update timers, a pattern that resembles https://github.com/temporalio/samples-go/blob/main/updatabletimer/workflow.go


examples/workflows/updatable_timer.rb
```diff
@@ -0,0 +1,37 @@
...
+      workflow.logger.info("Started timer with duration: #{sleep_duration}s")
+
+      workflow.wait_until do
+        timer.finished? || signal_received
+      end
+
+      if timer.finished?
+        workflow.logger.info("Timer fired")
+        break
+      end
```
JProgrammer, line 32: You could add

```
else
    timer.cancel
end
```

Issue #285, opened 2024-02-06
rubygems release 


edgarMeinart:
Please publish the current version (0.0.3) of this gem in rubygems.

https://rubygems.org/gems/temporal-ruby/versions/0.0.1

DeRauk:
@antstorm could you help out here? I believe you're still the sole ruby gems owner.

PR #305, opened 2024-07-11
Replicate official SDK to fetch results from Temporal

abhishekjain16:
While running long workflows that require synchronous results, `await_workflow_result` throws the following exception:
```
await_workflow_result': undefined method 'type' for nil:NilClass (NoMethodError)
```
pointing to line https://github.com/coinbase/temporal-ruby/blob/master/lib/temporal/client.rb#L251

Reason:
When trying to fetch workflow execution history with `event_type: :close`, Temporal sends empty events payload after next event (not necessarily close event)
Calling `.first.type` fails on empty event.

Fix:
Follow what official SDK does in Go and TS. Create a loop and try and fetch close event till it can find one
TS Example:
https://github.com/temporalio/sdk-typescript/blob/913e67e6381c421c3ed02ee24518adfa1243c84c/packages/client/src/workflow-client.ts#L630

Go Example:
https://github.com/temporalio/sdk-go/blob/626785769094562b39662d6076304c807b596ed9/internal/internal_workflow_client.go#L433-L436

This PR tries to replicate the same behavior

### DIFF
--- a/lib/temporal/client.rb
+++ b/lib/temporal/client.rb
@@ -229,6 +229,7 @@ module Temporal
       execution_options = ExecutionOptions.new(workflow, options, config.default_execution_options)
       max_timeout = Temporal::Connection::GRPC::SERVER_MAX_GET_WORKFLOW_EXECUTION_HISTORY_POLL
       history_response = nil
+      closed_event = nil
       loop do
         begin
           history_response = connection.get_workflow_execution_history(


lib/temporal/client.rb
```diff
@@ -229,25 +229,29 @@ def await_workflow_result(workflow, workflow_id:, run_id: nil, timeout: nil, nam
...
-          wait_for_new_event: true,
-          event_type: :close,
-          timeout: timeout || max_timeout,
-        )
-      rescue GRPC::DeadlineExceeded => e
-        message = if timeout
-          "Timed out after your specified limit of timeout: #{timeout} seconds"
-        else
-          "Timed out after #{max_timeout} seconds, which is the maximum supported amount."
+      while(true) do
```
bdchauvette, line 232: Just a suggestion to use some sugar:

```suggestion
      loop do
```

lib/temporal/client.rb
```diff
@@ -229,25 +229,29 @@ def await_workflow_result(workflow, workflow_id:, run_id: nil, timeout: nil, nam
...
+          else
+            "Timed out after #{max_timeout} seconds, which is the maximum supported amount."
+          end
+          raise TimeoutError.new(message)
         end
-        raise TimeoutError.new(message)
+        history = Workflow::History.new(history_response.history.events)
+        closed_event = history.events.first
+
+        break unless closed_event.nil?
```
bdchauvette, line 253: Totally a personal nitpick, but framing this in a positive way ("break if x is present") is easier to parse than "break unless x is nil / not present". This is just a personal thing, and the `unless` form might be totally more approachable for other folks.

I know that the suggest doesn't have quite the same semantics as the current code, — `break if closed_event` would break on _any_ non-nil value — but I think since it's coming from gRPC, we should be able to trust that the value we get back has the expected shape (as long as it's not `nil`)?

```suggestion
        break if closed_event
```

lib/temporal/client.rb
```diff
@@ -229,25 +229,29 @@ def await_workflow_result(workflow, workflow_id:, run_id: nil, timeout: nil, nam
...
+            event_type: :close,
+            timeout: timeout || max_timeout,
+          )
+        rescue GRPC::DeadlineExceeded => e
+          message = if timeout
+            "Timed out after your specified limit of timeout: #{timeout} seconds"
+          else
+            "Timed out after #{max_timeout} seconds, which is the maximum supported amount."
+          end
+          raise TimeoutError.new(message)
```
bdchauvette, line 248: If the plan is to bring this in line with the other SDKs, should this be a `retry` rather than a `raise`? Maybe with support for a max retry count that defaults to `1` for backwards compatibility?

abhishekjain16: TypeScript SDK does not have retries for timeout. I think it should be left to the client to decide on how they want to deal with it.

lib/temporal/client.rb
abhishekjain16: Done!

lib/temporal/client.rb
abhishekjain16: Done!

lib/temporal/client.rb
bdchauvette: I think it should follow the behavior of the other SDKs, but that could trip up any current code that assumes that calling `await_workflow_result` will succeed or time out.

Put another way, if code no longer times out, current users may end up with some nasty latency surprises. And since this gem's version doesn't get bumped (or officially published; #285) very often, I don't know if there'd be a good way to signal a big breaking change like that 😬

bdchauvette: Whoops, I get what you're saying now. I overlooked that the TS SDK also raises on deadline exceeded errors, it just hides it behind the `rethrowGrpcError` method 👍

morgan121:
I am also getting this error. Is it possible to merge this fix? It's causing us a bit of headache

abhishekjain16:
@cj-cb @DeRauk Do you think we can merge this? I can fix conflict but only if there a need to do it

PR #311, opened 2024-07-26
Support passing activity task rate limit on worker options

jazev-stripe:
### Summary

This PR exposes the `activity_max_tasks_per_second` option when constructing a new `Temporal::Worker` instance. This setting sets the `max_tasks_per_second` option when making `PollActivityTaskQueue` RPCs to the server, which has the effect of limiting how fast the server will give out activity tasks (both new activities and retries of already-running activities).

This same option is exposed in other SDKs; for example in Java it's available on the `WorkerOptions` struct as [the `setMaxTaskQueueActivitiesPerSecond` option](<https://javadoc.io/doc/io.temporal/temporal-sdk/latest/io/temporal/worker/WorkerOptions.Builder.html#setMaxTaskQueueActivitiesPerSecond(double)>). The doc comment added in this PR was based off of the doc comment in the Java SDK. The semantics of the values `activity_max_tasks_per_second` can be (non-negative decimal number) and when it sets set on the wire (if it's > 0) was copied directly from [the Java SDK's implementation, here](https://git.corp.stripe.com/stripe-private-oss-forks/temporal-sdk-java/blob/7e3902e3e7b59ee2c85415524eafcbdeccdc4f7d/temporal-sdk/src/main/java/io/temporal/internal/worker/ActivityPollTask.java#L73-L78) (although it diverges slightly from the [Go SDK](https://github.com/temporalio/sdk-go/blob/9c40461e456e36dc4e23215d73f410ef6e9b85dd/internal/internal_task_pollers.go#L943)).

### Motivation

Allow Ruby workers to control the rate of processing activities, to control impact on downstream systems.

#### Unit tests

Existing unit tests/example unit tests pass, and the new tests I added validate that, at each step, the option is making its way to the underlying gRPC connection:

```sh
bundle exec rspec spec/
cd examples && bundle exec rspec spec/
```

#### Manual tests

I also manually validated that the activity task rate limit works as expected. To do so, I modified the `examples/bin/worker` script to include a rate-limit on the `Worker.new` call:

```rb
worker = Temporal::Worker.new(binary_checksum: `git show HEAD -s --format=%H`.strip, activity_max_tasks_per_second: 0.5)
```

Then, I ran the `AsyncHelloWorldWorkflow` workflow with 10 activities, and validated that it took around 20-30 seconds (10 activities \* 0.5 activities/second + a bit of overhead/inaccuracy) for the workflow to complete:

```sh
cd examples
docker-compose up
bin/worker # with special code patch
bin/trigger AsyncHelloWorldWorkflow 10
```

image

Repeating the test _without_ the change to `examples/bin/worker` (i.e. an unlimited activity task queue) shows that the rate-limit makes a meaningful difference:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ Temporal::Worker.new(
   workflow_thread_pool_size: 10, # how many threads poll for workflows
   binary_checksum: nil, # identifies the version of workflow worker code
   activity_poll_retry_seconds: 0, # how many seconds to wait after unsuccessful poll for activities
-  workflow_poll_retry_seconds: 0  # how many seconds to wait after unsuccessful poll for workflows
+  workflow_poll_retry_seconds: 0,  # how many seconds to wait after unsuccessful poll for workflows
+  activity_max_tasks_per_second: 0 # rate-limit for starting activity tasks (new activities + retries) on the task queue
 )
 ```
 
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Temporal::Worker.new(
   workflow_thread_pool_size: 10, # how many threads poll for workflows
   binary_checksum: nil, # identifies the version of workflow worker code
   activity_poll_retry_seconds: 0, # how many seconds to wait after unsuccessful poll for activities
-  workflow_poll_retry_seconds: 0,  # how many seconds to wait after unsuccessful poll for workflows
+  workflow_poll_retry_seconds: 0, # how many seconds to wait after unsuccessful poll for workflows
   activity_max_tasks_per_second: 0 # rate-limit for starting activity tasks (new activities + retries) on the task queue
 )
 ```
--- a/lib/temporal/activity/poller.rb
+++ b/lib/temporal/activity/poller.rb
@@ -11,7 +11,8 @@ module Temporal
     class Poller
       DEFAULT_OPTIONS = {
         thread_pool_size: 20,
-        poll_retry_seconds: 0
+        poll_retry_seconds: 0,
+        max_tasks_per_second: 0 # unlimited
       }.freeze
 
       def initialize(namespace, task_queue, activity_lookup, config, middleware = [], options = {})
@@ -91,7 +92,8 @@ module Temporal
       end
 
       def poll_for_task
-        connection.poll_activity_task_queue(namespace: namespace, task_queue: task_queue)
+        connection.poll_activity_task_queue(namespace: namespace, task_queue: task_queue,
+                                            max_tasks_per_second: max_tasks_per_second)
       rescue ::GRPC::Cancelled
         # We're shutting down and we've already reported that in the logs
         nil
@@ -113,6 +115,10 @@ module Temporal
 
       def poll_retry_seconds
         @options[:poll_retry_seconds]
+      end
+
+      def max_tasks_per_second
+        @options[:max_tasks_per_second]
       end
 
       def thread_pool
--- a/lib/temporal/connection/grpc.rb
+++ b/lib/temporal/connection/grpc.rb
@@ -256,7 +256,7 @@ module Temporal
         client.respond_workflow_task_failed(request)
       end
 
-      def poll_activity_task_queue(namespace:, task_queue:)
+      def poll_activity_task_queue(namespace:, task_queue:, max_tasks_per_second: 0)
         request = Temporalio::Api::WorkflowService::V1::PollActivityTaskQueueRequest.new(
           identity: identity,
           namespace: namespace,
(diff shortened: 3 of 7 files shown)


jeffschoner (approved):

cj-cb (approved):

gytisgreitai:
@jazev-stripe can you elaborate a bit more of how this should work?

gytisgreitai:
@jazev-stripe can you elaborate a bit more of how this should work?

I've launched a single worker:
``` ruby
Temporal::Worker.new(
    activity_max_tasks_per_second: 0.1,
    activity_thread_pool_size: 1,
  )
```

and have one activity + one workflow:
```ruby
class HelloWorldWorkflow < Temporal::Workflow
  def execute
    HelloWorldActivity.execute!
  end
end

class HelloWorldActivity < Temporal::Activity
  def execute
    Net::HTTP.get_response(URI.parse('http://localhost:8000'))

    "Hello World"
  end
end
```

then I've locally launched a simple python server on the 8000 port. And instantly scheduled 20 workflows with simple `(1..20).each do ....`And what I'm seeing is a bit strange to me:
```
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:01] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:01] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:04] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:04] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:08] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:08] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:12] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:12] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:16] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:20] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:24] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:28] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:23:40] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:24:20] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:25:00] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:25:40] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:26:20] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:27:00] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:27:40] "GET / HTTP/1.1" 200 -
::ffff:127.0.0.1 - - [16/Aug/2024 15:28:20] "GET / HTTP/1.1" 200 -
```

I would somehow expect it to run once every 10 seconds? But what I get is two parallel activities every 4 seconds which seem to then slow down until almost 40 seconds between runs?

jazev-stripe:
> I would somehow expect it to run once every 10 seconds? But what I get is two parallel activities every 4 seconds which seem to then slow down until almost 40 seconds between runs?

The rate-limiting is done on the server-side, so unfortunately I think this is just a result of the rate-limiting not being super accurate for very small values. I'm not familiar with the server implementation, but if you ask in [the Temporal communnity Slack](https://temporal.io/community) they would probably be able to give you more details.

If I'm right, you should be able to reproduce that test/the results with any of the official SDKs, since they all use the same mechanism to implement activity task queue rate-limiting (that one Protobuf field when polling for activity tasks).

I also noticed similar phenomenon when testing locally ("+ a bit of overhead/inaccuracy"):

> it took around 20-30 seconds (10 activities * 0.5 activities/second + a bit of overhead/inaccuracy) for the workflow to complete...

One other thing to keep in mind is that the "current" value of the rate-limit seems to be stored on the server side in a way that takes O(minutes) for a new value to propagate. So, make sure you haven't run the worker on the same task queue without a rate-limit shortly before your test. In other words, make sure your test runs with a "clean slate" on that task queue, to make sure it's not being affected by anything else (you could try cleaning all of the server state before running your test to be sure).

PR #314, opened 2024-08-19
[Refactor] Remove Temporal::Concerns::Payloads

antstorm:
This PR is a resurrection of #152, that got too far behind the `HEAD`.

This is the first (but definitely the biggest by far) step to resolving #143. Since the introduction of `Temporal::Concerns::Payloads` everything that was using it was dependent on the primary singleton configuration. This means that you can't have 2 clients or workers within the same app that are using different configs. The solution is explicitly injectable configuration.

This PR introduces `Temporal::ConverterWrapper` that has all the functionality of the `Temporal::Concerns::Payloads`, but is meant to be provided explicitly to the parts of the code that need it. This way there are no hidden dependencies on the global configuration.

Once this is merged, a small number of references to `Temporal.configuration` will need to be removed and the deprecation warning should go away for good.


antstorm:
Here's the fix for the failing integration specs — #315

ukd1:
@antstorm there are some fails on the test_examples fyi!

antstorm:
@ukd1 yep, these should be fixed by https://github.com/coinbase/temporal-ruby/pull/315

cj-cb (approved):

antstorm:
@cj-cb rebased with master without any changes, this should now be good to go

PR #343, opened 2025-06-28
Remove deprecated `Temporal.configuration` function

### LINKED ISSUE #143: Deprecation message spam

aguynamedben:
Fixes #143. For a few years there has been a noise `warn` message telling people to not use the function Temporal.configuration, but this function was still used internall in temporal-ruby.

The options were:
a) Remove the logger warn message, as whoever in the past had planned to deprecate access via `Temporal.configuration` never followed up. I don't know why they were trying to remove this method of accessing it. b) **[CHOSEN OPTION]** Raise the `config` function from `private` to be accessible externally. I don't see why this wouldn't be okay.

I ran the tests per the instructions and I get the same amount of test passes, without the noisy deprecation warnings.

I don't know the full history that led to this being half-deprecated, but this seems like a fine way to clean it up.

## Before
image

## After
image

## Breaking Change

CHANGELOG.md should have:

* **Breaking change:** `Temporal.configuration` was removed and replace by `Temporal.config`. If you previously relied on `Temporal.configuration` just update the reference to `Temporal.config`.

Also if this Gem is still doing any versioning, it's a MAJOR patch (i.e. first digit) because if removes a public-facing API, i.e. `Temporal.configuration`. If we don't want to create a breaking change, I recommend Option A above, just remove the annoying log message and let the method live on into the future.

### DIFF
--- a/lib/temporal.rb
+++ b/lib/temporal.rb
@@ -56,9 +56,8 @@ module Temporal
       @default_client = nil
     end
 
-    def configuration
-      warn '[DEPRECATION] This method is now deprecated without a substitution'
-      config
+    def config
+      @config ||= Configuration.new
     end
 
     def logger
@@ -75,9 +74,6 @@ module Temporal
       @default_client ||= Client.new(config)
     end
 
-    def config
-      @config ||= Configuration.new
-    end
 
   end
 end
--- a/lib/temporal/testing/local_workflow_context.rb
+++ b/lib/temporal/testing/local_workflow_context.rb
@@ -12,7 +12,7 @@ module Temporal
     class LocalWorkflowContext
       attr_reader :metadata, :config
 
-      def initialize(execution, workflow_id, run_id, disabled_releases, metadata, config = Temporal.configuration)
+      def initialize(execution, workflow_id, run_id, disabled_releases, metadata, config = Temporal.config)
         @last_event_id = 0
         @execution = execution
         @run_id = run_id
--- a/lib/temporal/testing/replay_tester.rb
+++ b/lib/temporal/testing/replay_tester.rb
@@ -12,7 +12,7 @@ module Temporal
     end
 
     class ReplayTester
-      def initialize(config: Temporal.configuration)
+      def initialize(config: Temporal.config)
         @config = config
       end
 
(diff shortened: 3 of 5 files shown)


lib/temporal.rb
```diff
@@ -75,9 +74,6 @@ def default_client
       @default_client ||= Client.new(config)
     end
 
-    def config
-      @config ||= Configuration.new
-    end
```
aguynamedben, line 80: Not removed, hoisted out of `private` ⬆️

aguynamedben:
Bump! @cj-cb

lib/temporal.rb
```diff
@@ -56,9 +56,8 @@ def configure(&block)
       @default_client = nil
     end
 
-    def configuration
-      warn '[DEPRECATION] This method is now deprecated without a substitution'
-      config
+    def config
+      @config ||= Configuration.new
     end
```
flavio-pego-smartfit: We have to figure out the history and motivation for `configuration` to be flagged as deprecated.

If you _hoist_ `config` from private, but still leave `configuration` here, it wont break other people's codes, right?

PR #349, opened 2025-11-20
[WorkerPoller] Make poller threads configurable

1pkg:
Add support for configurable number of poller threads via `activity_poller_threads` and `workflow_poller_threads` config.


spec/fabricators/grpc/history_event_fabricator.rb
1pkg, line 1: Fixes broken tests

PR #350, opened 2025-11-27
cache default identity

1pkg:


lib/temporal/configuration.rb
```diff
@@ -135,7 +135,7 @@ def default_identity
       hostname = `hostname`
       pid = Process.pid
 
-      "#{pid}@#{hostname}".freeze
+      @default_identity ||= "#{pid}@#{hostname}".freeze
```
d2army: That `hostname` system call still gets called every time default_identity gets called?

What if you replace default_identity like this:

```
def default_identity
  @default_identity ||= begin
    hostname = `hostname`      #  now, this for sure runs once
    pid = Process.pid          # now, this for sure runs once
    "#{pid}@#{hostname}".freeze
  end
end

```

1pkg: good catch!

PR #355, opened 2026-05-04
 Ignore unsupported and Nexus events

jeffschoner:
### Summary

- Add graceful handling for event types the Ruby SDK doesn't support (Nexus operations, `WORKFLOW_EXECUTION_OPTIONS_UPDATED`). These events are logged and skipped rather than crashing the worker.
- Honor the `worker_may_ignore` bit on history events, allowing the server to introduce new event types without breaking older workers.
- Upgrade Temporal API proto definitions to v1.62.9, which includes enum values for Nexus and other newer event types.
- Add a replay test verifying workflows replay correctly when histories contain ignored events.
- Fix pre-existing test failures: GRPC poll mock keyword arg mismatch and a flaky clock-skew assertion in the metadata integration spec.

### Details

When a workflow history contains event types the SDK doesn't handle (e.g. Nexus operations or `WORKFLOW_EXECUTION_OPTIONS_UPDATED`), the state manager now:
1. Checks against a known list of unsupported-but-ignorable event types and skips them with a warning log.
2. For truly unknown event types not in the list, checks the `worker_may_ignore` flag from the server — if set, logs and skips; otherwise raises `UnsupportedEvent`.

The proto upgrade to v1.62.9 ensures these event types are properly deserialized (previously they mapped to `EVENT_TYPE_UNSPECIFIED` since the enum didn't include them).

### Test plan

- [x] `bundle exec rspec` — 662 examples, 0 failures
- [x] `examples/spec/replay/signal_with_start_ignored_event_spec.rb` — replay test passes with ignored eve


jazev-stripe (approved):
The Ruby changes all match what we are running at Stripe in our fork (and have successfully deployed to our services).

The Protobuf generation changes should be good; in particular the `add_serialized_file` API used by more modern protoc is compatible going back to `google-protobuf` v3.18: [source](https://protobuf.dev/news/2023-04-20/#backward)

